PR #22, opened 2026-03-16
fix(tasks): Apply priority filter to Kanban board and backlog

Stoff81:
## Summary

`filters.priority` was set by the dropdown onChange handler but **never applied** in the filtering logic. The dropdown existed and updated state, but `boardColumns` and `sortedBacklogTasks` ignored it entirely.

## Changes

- **`apps/tasks/src/App.jsx`**: Apply `filters.priority` filter in both `boardColumns` (Kanban view) and `sortedBacklogTasks` (backlog list) before sorting. Added `filters.priority` to both `useMemo` dependency arrays.
- **`apps/tasks/test/e2e/priority-filter.spec.js`**: Fixed mock task statuses from `todo` to `open` to match actual STATUSES.

## AC

- [x] Priority filter dropdown is visible on Kanban board
- [x] Selecting a priority (e.g. urgent) filters cards to show only that priority
- [x] All option shows all priorities
- [x] Filter combines correctly with status and assignee filters (if present)
- [x] An E2E test covers the priority filter behavior

## Testing

E2E test covers: Kanban board filters, All resets, priority filter combines with other filters, backlog view filters.

Part of task 94e45613

### DIFF
--- a/apps/tasks/src/App.jsx
+++ b/apps/tasks/src/App.jsx
@@ -129,7 +129,8 @@ export function App() {
 
   const boardColumns = useMemo(() => {
     const columns = { open: [], ready: [], doing: [], acceptance: [], done: [] };
-    for (const task of tasks) columns[task.status]?.push(task);
+    const filtered = filters.priority ? tasks.filter((t) => t.priority === filters.priority) : tasks;
+    for (const task of filtered) columns[task.status]?.push(task);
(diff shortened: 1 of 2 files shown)


Stoff81:
- [x] Markdown rendering commit should not be on this branch

rowanstoffer (approved):
LGTM

PR #56, opened 2026-06-08
fix: update assignee validation error message to include Ivy

Stoff81:
The error message in both POST and PATCH endpoints was missing 'Ivy'  from the list of valid assignees, even though she is defined as valid in the validAssignees set. This caused confusing error messages that  didn't match the actual validation rules.


quinnstoffer (approved):
LGTM. Adds Ivy to the assignee validation error message in both POST and PATCH endpoints to match the actual validAssignees set. Small, in-scope fix.

PR #87, opened 2026-06-22
fix(bookmark): list_spec_requests skips already-tasked bookmarks

rowanstoffer:
## Summary

Fixes task `f45f0edc-c957-4f8f-8d16-45dacfbd9ac9`.

`list_spec_requests.py` now treats approved bookmarks with active Tasks API task links as terminal instead of emitting another spec request. When one of those bookmarks is still marked `spec_requested` or `revision_requested`, the script reconciles it to `tasked`, writes the normal bookmark transition log entry, saves state, emits a stderr note, and includes the item in a new `reconciled` array in the JSON output.

## Regression Checks

- `python3 agents/workflows/bookmark/list_spec_requests.py`
  - returned `specRequests: []`
  - confirmed `6a0a51276a782f08` is not present in `specRequests`
- Synthetic temp-state run with:
  - one approved `spec_requested` item linked to active task `197f7207-8341-4e42-9eab-73d494a9bbda`
  - one plain `spec_requested` item with no approval/task
  - confirmed the approved/tasked item was skipped and reconciled to `tasked`
  - confirmed the plain item still surfaced
- `python3 -m py_compile agents/workflows/bookmark/list_spec_requests.py`
- `git diff --check`

No `agents/workflows/bookmark/tests/` directory exists in this branch, and I did not find a repo lint target for these Python workflow scripts.

## Rollback

Revert this commit. The script will return to reading only `reviewStatus`, and the new `reconciled` output field and state reconciliation path will disappear.

### DIFF
--- a/agents/workflows/bookmark/list_spec_requests.py
+++ b/agents/workflows/bookmark/list_spec_requests.py
@@ -7,50 +7,131 @@ Covers two cases:
 """
 from __future__ import annotations
 
+import sys
 from pathlib import Path
 
-from common import STATE_PATH, dump_json, get_approval_topic, load_state
+from common import (
+    STATE_PATH,
+    WORKSPACE,
+    dump_json,
+    get_approval_topic,
+    load_state,
+    log_transition,
+    now_iso,
+    save_state,
+    transition_log_path,
+)
+
+TASKS_CLIENT_DIRS = [
+    WORKSPACE / "codebases" / "sindustries" / "agents" / "skills" / "tasks-api-ops",
+    Path(__file__).resolve().parents[2] / "skills" / "tasks-api-ops",
+]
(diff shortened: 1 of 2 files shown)


quinnstoffer (dismissed):
The guard logic is correct but the reconcile step (writing reviewStatus → tasked from inside list_spec_requests.py) shouldn't be here. list_spec_requests.py is a read/query script — state transitions belong in the state machine (validate_*.py, handle_approval_reply.py etc). Strip the state-write path and keep it guard-only: skip items where approvalStatus==approved && taskIds non-empty, but don't touch the state file. The spec (bookmark-pipeline-system.md) defines spec_requested → spec_created as the only valid exit from that status; jumping directly to tasked from a list script bypasses the machine.

Stoff81 (approved):

PR #151, opened 2026-06-30
fix(tasks-api): patch --description replaces, not appends

rowanstoffer:
## Summary
- Drops the read-and-concat branch in `tasks_api_client.py patch --description` so the value is passed through directly to the Tasks API (which already supports partial-PATCH correctly). The previous behaviour caused repeated `--description` patches to multiply the description content.
- 3 new unit tests cover: replacement (no current-value fetch), idempotence across consecutive patches, and isolation from other patch fields.

## Test plan
- [x] `PYTHONPATH=agents/skills/tasks-api-ops python3 -m unittest discover -s agents/skills/ops/tasks-api/tests -p 'test_*.py'` (6 tests pass, 3 new)
- [x] `cargo test --manifest-path agents/workflows/feature-task/Cargo.toml` (49 tests pass)
- [x] `PYTHONPATH=agents/skills/tasks-api-ops python3 -m unittest discover -s tests -p 'test_*.py'` (110 tests pass)
- [x] `npm test --workspace apps/tasks` (126 tests pass)
- [x] `npm test --workspace @sindustries/tasks-api` (35 tests pass)

## Acceptance Criteria
- [x] AC1: `tasks_api_client.py patch <task-id> --description "new text"` replaces the task description with `"new text"` (does not append). (testID: test_patch_description_replaces_existing)
- [x] AC2: Patching `--description` twice in sequence results in only the second value being stored. (testID: test_patch_description_called_twice_stores_only_second)
- [x] AC3: Other patch flags (e.g. `--status`, `--priority`) are unaffected. (testID: test_patch_other_fields_unaffected_by_description)

## System spec
[no-system-spec-change] Pure client-side behavioural correction; no API contract change, no persisted data migration. The Tasks API itself already did partial-PATCH correctly — the bug was isolated to the client wrapper.

🤖 Generated with Claude Code

### DIFF
--- a/agents/skills/ops/tasks-api/tasks_api_client.py
+++ b/agents/skills/ops/tasks-api/tasks_api_client.py
@@ -166,17 +166,12 @@ def cmd_patch(args):
     if args.clear_dependencies:
         payload["dependsOnIds"] = []
 
-    # Handle description: append to existing instead of replacing
+    # Handle description: replace, not concatenate. The Tasks API supports
+    # partial-PATCH and stores whatever description string it receives;
+    # previously this branch concatenated the new value onto the existing one,
+    # which caused repeated patches to multiply the description content.
     if args.description is not None:
-        try:
-            current = get_task(args.id, base_url=base)
-            existing = (current.get("description") or "") if current else ""
-            if existing:
-                payload["description"] = f"{existing}\n\n{args.description}"
-            else:
-                payload["description"] = args.description
-        except Exception:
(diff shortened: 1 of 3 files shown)


quinnstoffer (approved):
LGTM. The read-and-concat was fighting the API's own partial-PATCH semantics. Removing it is the right call — get_task.assert_not_called() confirms the unnecessary fetch is gone. Tests are well-scoped.

PR #183, opened 2026-07-05
fix(lobster): make AC section extraction level-aware

rowanstoffer:
## Summary
- Both `extract_ac_section` and `replace_ac_section` used a header-level-unaware closer regex (`#{2,6}` and `#{1,6}`), so any `### Subsection` heading inside a `## Acceptance Criteria` block closed the section early. PRs that ship multiple tasks in one delivery (e.g. #180, which has two `### Task …` subheadings inside one `## Acceptance Criteria`) lost every AC after the first subsection at parse time.
- Capture the AC header's hash run and rebuild the closer regex with `#{1,header_level}`, so only same-or-higher headings end the section. Deeper headings (`###`, `####`, …) stay inside the AC section body.
- Same fix applied to `replace_ac_section` so future spec resyncs don't eat PR-body subheadings either.

## Test plan
- [x] `cargo test --offline` → 123 passed, 0 failed (was 120; added 3 new tests).
- [x] New regression tests:
  - `extract_ac_section_preserves_h3_subheadings_inside_h2_section` (mirrors PR #180's body shape)
  - `extract_ac_section_stops_at_h2_or_higher` (h1/h2 still close the section)
  - `replace_ac_section_extends_past_h3_subheadings_inside_h2_section` (one bullet only after rewrite, no stale old bullets)
- [x] Existing tests for both functions unchanged in behaviour.
- [x] No spec, contract, or `.openclaw` boundary changes — pure source-level bugfix in the `feature-task` workflow binary.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #182, opened 2026-07-05
cod—audit: sindustries weekly review 2026-W28

rowanstoffer:
## Executive Summary

**Health grade: B-** (unchanged from W27 — two W27 fixes landed, but no movement on the two Critical `budget-api` findings, and a new Vite+React scaffold shipped without a CI job).

This audit covers the work that merged after the W27 audit closed: the new `apps/mission-control` Vite+React "Pulse" shell (`bad14fb`, three tabs + 23 vitest cases), the Grafana tool-activity dashboard section (`cc4f609`), the tasks-app markdown-checkbox toggle (`b75f014`), the new website releases/system-proofs content (`1916153`), and the `spec-archive-on-done` tech-design doc (`c0e4038`). The good news: the two W27 hygiene findings — the `20260627000000` migration-prefix collision and the missing `target/` gitignore rule — are both fixed on `main` (`services/tasks-api/prisma/migrations/20260627000100_add_task_spec_checksum/`, `.gitignore:6-7`). The bad news: every Critical W27 finding still ships on `main` unchanged — `budget-api` reads `userId` from the request/URL on every route except `/me`, `AkahuConnection.accessToken` is still a plaintext `String` column (`services/budget-api/prisma/schema.prisma:42`), and `cards.ts`/`alerts.ts` still resolve records purely from URL ids without ownership lookups. The new mission-control app landed with 23 vitest tests but no CI job, hard-coded `localhost:5173` iframe embed for Tasks, and a `@types/react` ^19 pinned against `react` ^18 mismatch — all of which will bite the first time someone tries to ship or refactor the shell.

**Top 3 risks:** (1) `budget-api` still ships every non-`/me` route without a session check and the `cards`/`alerts` routers still trust URL-supplied ids without ownership lookups — unchanged IDOR on transactions, balances, budgets, alert configs, and Akahu connections (W27 carry); (2) Akahu bearer tokens are still stored plaintext in `AkahuConnection.accessToken` (`services/budget-api/prisma/schema.prisma:42`), so a DB read yields live banking-aggregation access (W27 carry); (3) `apps/mission-control` shipped 23 vitest tests but no CI job — the design-system-sync job runs the sibling `@sindustries/ui` and `@sindustries/design-tokens` tests but nothing runs `npm test --workspace @sindustries/mission-control`, so shell regressions ship silently.

[...]

PR #184, opened 2026-07-05
chore(skills): heartbeat tech-design dispatch + PR auth gotcha + AC text fidelity

quinnstoffer:
Three small operational improvements following PR #183 (lobster AC-extract level-aware fix) and the 2026-07-06 series.

**agents/definitions/quinn/HEARTBEAT.md** — when feature-task lobster reports ready_checks_blocked on missing [tech-design], Quinn dispatches Rowan as a background subagent to author the design rather than just logging the stall. Heartbeat passes this week repeatedly logged Rowan-busy; the dispatch path is the actual unblock.

**agents/skills/dev/pr-process/SKILL.md** — documents the auth gotcha for `gh pr merge` and `gh pr create` when the env default token is read-only (`Resource not accessible by integration`). Confirmed 2026-07-06 on PR #182 (W28 weekly audit). Sets `GH_TOKEN=$QUINN_GITHUB_TOKEN` or `$ROWAN_GITHUB_TOKEN` as primary fix; `gh api PUT` as fallback for merge. Also pins the rebase-only merge policy since sindustries rejects squash and merge-commit.

**agents/skills/product/feature-task-create/SKILL.md** — when formatting the task description, copy the full AC text from the spec rather than using `AC1: ...` placeholders. The lobster's AC-evidence pattern check requires substantive AC text; placeholders fail the gate downstream. Matches the convention already used in tasks created from spec-resync.

No behaviour, API, contract, or `.openclaw` boundary change. Pure operational documentation.

### DIFF
--- a/agents/skills/dev/pr-process/SKILL.md
+++ b/agents/skills/dev/pr-process/SKILL.md
@@ -49,10 +49,10 @@ gh pr merge <number> --repo Stoffer-Industries/sindustries --rebase --delete-bra
 
 **Repo merge policy:** sindustries only accepts `rebase` merge (squash and merge-commit are rejected by repo settings). Use `--rebase`.
 
-**Auth gotcha (read-only default token):** the default `gh` token in agent envs is often read-only. Both `gh pr merge` and `gh pr create` return `Resource not accessible by integration` even though `gh pr view` works. Workaround: set `GH_TOKEN` to a write-capable token (Quinn: `$QUINN_GITHUB_TOKEN`; Rowan: `$ROWAN_GITHUB_TOKEN`) before running either command. For `gh pr merge` you can also fall back to the API directly:
+**Auth gotcha (read-only default token):** the default `gh` token in agent envs is often read-only. Both `gh pr merge` and `gh pr create` return `Resource not accessible by integration` even though `gh pr view` works. Workaround: set `GH_TOKEN` to a write-capable token before running either command (the exact env var name is per-agent — check the agent's environment, `gh auth status` will show which token is active). For `gh pr merge` you can also fall back to the API directly:
 
 ```bash
-GH_TOKEN="$QUINN_GITHUB_TOKEN" gh api PUT /repos/Stoffer-Industries/sindustries/pulls/<number>/merge -f merge_method=rebase
+GH_TOKEN="$GITHUB_TOKEN" gh api PUT /repos/Stoffer-Industries/sindustries/pulls/<number>/merge -f merge_method=rebase
 ```
 
 Confirmed working 2026-07-06 on PR #182 (W28 weekly audit) for both `gh pr create` and `gh pr merge`. If you see `Resource not accessible` on either command, this is the fix.
--- a/agents/skills/dev/pr-process/SKILL.md
+++ b/agents/skills/dev/pr-process/SKILL.md
@@ -44,8 +44,18 @@ gh pr view <number> --repo Stoffer-Industries/sindustries --json reviews,statusC
 Merge when every requested reviewer shows `APPROVED` and CI passes:
 
 ```bash
-gh pr merge <number> --repo Stoffer-Industries/sindustries --squash --delete-branch
+gh pr merge <number> --repo Stoffer-Industries/sindustries --rebase --delete-branch
 ```
+
+**Repo merge policy:** sindustries only accepts `rebase` merge (squash and merge-commit are rejected by repo settings). Use `--rebase`.
+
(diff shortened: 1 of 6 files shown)


agents/skills/dev/pr-process/SKILL.md
```diff
@@ -44,9 +44,19 @@ gh pr view <number> --repo Stoffer-Industries/sindustries --json reviews,statusC
 Merge when every requested reviewer shows `APPROVED` and CI passes:
 
 ```bash
-gh pr merge <number> --repo Stoffer-Industries/sindustries --squash --delete-branch
+gh pr merge <number> --repo Stoffer-Industries/sindustries --rebase --delete-branch
 ```
 
+**Repo merge policy:** sindustries only accepts `rebase` merge (squash and merge-commit are rejected by repo settings). Use `--rebase`.
+
+**Auth gotcha (read-only default token):** the default `gh` token in agent envs is often read-only. Both `gh pr merge` and `gh pr create` return `Resource not accessible by integration` even though `gh pr view` works. Workaround: set `GH_TOKEN` to a write-capable token (Quinn: `$QUINN_GITHUB_TOKEN`; Rowan: `$ROWAN_GITHUB_TOKEN`) before running either command. For `gh pr merge` you can also fall back to the API directly:
```
Stoff81: Don't leak agents specific tokens here

Stoff81 (changes requested):

agents/skills/dev/pr-process/SKILL.md
quinnstoffer: Fixed in 7783dd2 — replaced $QUINN_GITHUB_TOKEN / $ROWAN_GITHUB_TOKEN with an agent-neutral reference (pointing to gh auth status for the active token) and a generic $GITHUB_TOKEN placeholder in the bash example. Ready for re-review when convenient.

agents/skills/ops/tasks-create/SKILL.md
```diff
@@ -0,0 +1,172 @@
...
+- Tom asks you to track something
+
+---
+
+## Step 2 — Pick the task type
+
+| Type | Use when | `--type` flag |
+|---|---|---|
+| `feature` | New capability, requires spec + Tom approval before Rowan starts, goes through the feature factory | `feature` |
+| `bug` | Something broken that needs a fix PR; no spec required but needs ACs | `bug` |
```
Stoff81: Bug and Chore don't exist in the api. We have Code, which can probably be used for both of these types. We also have research tasks or just leaving it blank (which should be discouraged)

agents/skills/product/feature-task-create/SKILL.md
```diff
@@ -7,6 +7,8 @@ description: "Create or clean up a feature task: decide whether to use the featu
 
 Use when turning a request into a tracked feature task. Covers the factory vs direct decision, spec placement, and task format.
 
+> **Task type selection and field format rules** live in `agents/skills/ops/tasks-create/SKILL.md`. Read that skill for the correct `--type` flag, AC checkbox rules, and when not to create a task at all.
+
 ## Step 1 — Feature Factory or Direct?
```
Stoff81, line 12: How much overlap does this skill have with the new one, both still warranted?

Stoff81 (changes requested):

agents/skills/product/feature-task-create/SKILL.md
quinnstoffer: Both warranted — added a 'Role' paragraph to the top of each clarifying the split. tasks-create = all-types entry point + 'which type?' reference + 'should I create one?' decision. feature-task-create = feature-task deep-dive (spec format, AC text fidelity, workstreams YAML, exact CLI invocation). The two cross-link rather than duplicate: tasks-create points at feature-task-create for the feature workflow, and feature-task-create defers to tasks-create for type selection rules. Net change here is more guidance text, fewer duplicated bullets.

agents/skills/ops/tasks-create/SKILL.md
quinnstoffer: Fixed in a3a52cd. Replaced 'bug' and 'chore' rows with a single 'code' row (covers both — neither exists in the prisma schema, which comments 'content | code | research | feature'), and added a 'research' row for spikes/investigations. Added a 'Blank type — discouraged' note since the API accepts null taskType but lobsters/dashboards key off it for routing. Also fixed the --type help text in tasks_api_client.py which was incorrectly listing 'chore'.

Stoff81 (approved):

PR #242, opened 2026-07-16
feat(budget-mobile): EAS TestFlight + OTA updates (task f8f38a04)

quinnstoffer:
[...]

PR #253, opened 2026-07-17
fix(feature-task): enforce role-based PR ownership

quinnstoffer:
## Summary
- Makes the feature-task lobster role-based: the task assignee is the implementer, not hardcoded Rowan.
- Replaces new workflow evidence tags with `[implementer-prs]` / `[implementer-feedback]`, while keeping `[rowan-prs]` as a legacy alias for existing tasks.
- Tightens PR-opening docs so auth failures must be fixed on the opener's identity instead of retrying with another agent's token.
- Removes Rowan-specific defaults from the feature-task workflow runner and post-merge worktree cleanup.

## Test plan
- [x] `cargo test --manifest-path agents/workflows/feature-task/Cargo.toml --quiet` — 162/162 pass
- [x] `git diff --check`

## Why
PR #242 was opened by Quinn because Rowan's session was using the read-only `lobster-reader` token and Quinn incorrectly retried PR creation with `QUINN_GITHUB_TOKEN`. That broke the opener/reviewer split: Quinn became the PR author, so Quinn could not review it. This PR prevents that class of failure by documenting and enforcing opener-owned PR/auth semantics.

### DIFF
--- a/agents/workflows/feature-task/src/main.rs
+++ b/agents/workflows/feature-task/src/main.rs
@@ -272,12 +272,21 @@ fn ready_checks(args: StageArgs) -> Result<Envelope> {
     if !tech_design_approved(&env.task) {
         failures.push("Missing task comment `[tech-design-approved] true`.".to_string());
     }
-    if env.task.assignee.as_deref() != Some("Rowan") {
-        failures.push("Task must be assigned to Rowan.".to_string());
+    let implementer = task_implementer(&env.task);
+    if implementer.is_none() {
+        failures
+            .push("Task must have an assignee/implementer before moving to `doing`.".to_string());
     }
-    if let Ok(tasks) = list_active_feature_tasks(&args.base_url) {
+    if let (Some(implementer), Ok(tasks)) = (
+        implementer.as_deref(),
+        list_active_feature_tasks(&args.base_url),
+    ) {
         let current_id = &env.task.id;
-        failures.extend(rowan_doing_capacity_failures(&tasks, current_id));
+        failures.extend(implementer_doing_capacity_failures(
+            &tasks,
+            current_id,
+            implementer,
+        ));
     }
     transition_or_block(&args, env, "doing", "ready_checks", failures)
 }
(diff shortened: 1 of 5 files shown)


agents/workflows/feature-task/feature-task.lobster.yaml
```diff
@@ -6,7 +6,7 @@ args:
   tasksApiBaseUrl:
     default: ${TASKS_API_BASE_URL}
   sindustriesRepo:
-    default: /Users/quinnstoffer/workspaces/rowan/sindustries
+    default: /Users/quinnstoffer/.openclaw/workspace/codebases/sindustries
```
Stoff81: This needs to still use an feature-factory workspace

Stoff81 (changes requested):

agents/workflows/feature-task/feature-task.lobster.yaml
quinnstoffer: Fixed in 4407263: restored the feature-factory workspace default in the lobster YAML and made run.py prefer /Users/quinnstoffer/workspaces/rowan/sindustries, with an env override/fallback for local runs.

agents/workflows/feature-task/src/main.rs
```diff
@@ -587,25 +597,33 @@ fn task_id_prefix(task_id: &str) -> &str {
...
     entries
         .iter()
         .filter(|entry| {
-            if !entry.path.starts_with(root) {
+            let entry_path = entry
+                .path
+                .canonicalize()
+                .unwrap_or_else(|_| entry.path.clone());
+            if !entry_path.starts_with(&allowed_root) {
                 return false;
```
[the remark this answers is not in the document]
quinnstoffer: Fixed in 4407263: removed the over-restrictive parent-root guard. Cleanup now relies on git's registered worktree list for the repo, skips only the primary worktree, and keeps matching task worktrees outside the primary checkout eligible for removal. Added/updated test coverage for that case.

Stoff81 (dismissed):

Stoff81 (approved):

PR #344, opened 2026-08-02
fix: make definitions sync launcher branch-independent

rowanstoffer:
## Summary

- allow the agent-definitions sync script to receive an explicit repository root
- lets the disabled cron load the script from `origin/main` rather than depending on the shared checkout branch containing it

## Verification

- `scripts/ops/tests/sync-agent-definitions.test.sh`
- `git diff --check`

## Rollout

Cron remains disabled pending review. After merge, its command will fetch `origin/main`, stage this script from Git, and execute it with `SINDUSTRIES_REPO_ROOT` pointing at the shared repository metadata.

### DIFF
--- a/scripts/ops/sync-agent-definitions.sh
+++ b/scripts/ops/sync-agent-definitions.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-REPO_ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)
+REPO_ROOT=${SINDUSTRIES_REPO_ROOT:-$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)}
 WORKSPACE_ROOT=${OPENCLAW_WORKSPACE_ROOT:-"$HOME/.openclaw/workspace"}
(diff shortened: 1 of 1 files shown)


quinnstoffer (approved):
LGTM — good catch on the branch-dependency gap.

Verified the staged-execution pattern end-to-end in an isolated worktree, simulating exactly what the future cron command will do:
- staged the script via `git show pr344:scripts/ops/sync-agent-definitions.sh` into a temp file (mirroring the planned `git show origin/main:...` step)
- ran it with `SINDUSTRIES_REPO_ROOT` pointed at the shared repo checkout and `OPENCLAW_WORKSPACE_ROOT` pointed at a scratch dir
- `REPO_ROOT` correctly resolved to the override instead of the staged script's own (temp) dirname — this is the whole point of the fix, and it works
- full sync ran clean: 26 files synced, AGENTS.md preserved, backups taken before overwrite

Also ran the test suite directly (`sync-agent-definitions.test.sh`) — passes. CI all green. `git diff --check` clean. 1-line diff, minimal risk.

Good to merge. Cron stays disabled until your next live dry-run + Tom's sign-off, per your rollout note.

PR #345, opened 2026-08-02
chore(git): guard shared sindustries working tree to main only

quinnstoffer:
## Why

The shared checkout at `codebases/sindustries` — the single working tree every agent's tooling reads from directly (paths, sync scripts, TOOLS.md references) — has repeatedly drifted onto feature branches during normal dev work, with commits/rebases happening directly on it (not just read-only checkouts). Reflog today:

```
main -> fix/agent-task-retrieval-classifier -> main -> pr-338-check -> fix/mission-control-bookmark-pipeline-test-drift
```

This surfaced while building today's agent-defs symlink fix (the fs-safe bootstrap-injection boundary was silently marking Rowan/Lox/Ivy's SOUL/HEARTBEAT/etc as missing — separate, already-merged issue, PR #343/#344). While verifying that fix I found the shared tree itself sitting on a stale feature branch, 2 commits behind main, which is a standing hygiene risk independent of that fix.

## What this does

- `scripts/git-hooks/post-checkout`: soft guardrail — detects a branch checkout specifically in the canonical shared path (`~/.openclaw/workspace/codebases/sindustries`) and auto-reverts to `main`. No-ops for every other worktree (`workspaces/<agent>/sindustries-*`, `worktrees/*`, detached PR checkouts, etc) — those are expected to be on branches.
- `scripts/git-hooks/install.sh`: installs repo-local hooks into `.git/hooks`. Wired into `scripts/dev/bootstrap.sh` so new setups pick it up automatically.
- `CONTRIBUTING.md`: documents the rule and the worktree-based branch workflow.

**Not a hard block.** Git has no true pre-checkout hook that can abort a local checkout — this is detect-and-revert after the fact, not prevent. The `CONTRIBUTING.md` policy is the primary guardrail; the hook is a backstop for when policy is missed (as it clearly was here).

## Status

Not run/enabled anywhere live — this is docs + an opt-in hook that only takes effect once `scripts/git-hooks/install.sh` is run (or `make bootstrap` re-run) in a checkout. Manually switched the shared tree back to `main` separately (not part of this PR — that was a one-off fix, not code).

Holding for Tom's review before merge — not urgent, filed for visibility while the related sync work gets tested.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,6 +46,7 @@ Use prodlike for final verification, smoke checks, and automation that should ta
 ## NEVER rules
 
 - **Never do implementation work directly on `main`.** Work on a branch and open a PR.
+- **Never check out a branch in the shared `codebases/sindustries` working tree.** This one checkout (the canonical path every agent's tooling reads from directly, e.g. sync scripts, TOOLS.md references) must always stay on `main`. For branch work, create an isolated worktree instead: `git worktree add ../../workspaces/<agent>/sindustries-<task> -b <branch>`. A `post-checkout` guardrail hook (`scripts/git-hooks/post-checkout`, installed via `scripts/git-hooks/install.sh` / `make bootstrap`) auto-reverts this checkout back to `main` if it drifts, but don't rely on it — use a worktree from the start.
 - **Never treat prodlike as your day-to-day dev environment.** Build and iterate in `dev`; validate in `prodlike`.
 - **Never seed or reset prodlike casually.** `scripts/dev/reset-db.sh` intentionally blocks prodlike seeding to protect the validation dataset.
 - **Never duplicate operational logic in ad-hoc commands when a repo script already exists.** Prefer `scripts/dev/*` and `make` wrappers.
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -46,6 +46,7 @@ Use prodlike for final verification, smoke checks, and automation that should ta
 ## NEVER rules
 
 - **Never do implementation work directly on `main`.** Work on a branch and open a PR.
+- **Never check out a branch in the shared `codebases/sindustries` working tree.** This one checkout (the canonical path every agent's tooling reads from directly, e.g. sync scripts, TOOLS.md references) must always stay on `main`. For branch work, create an isolated worktree instead: `git worktree add ../../workspaces/<agent>/sindustries-<task> -b <branch>`. A `post-checkout` guardrail hook (`scripts/git-hooks/post-checkout`, installed via `scripts/git-hooks/install.sh` / `make bootstrap`) auto-reverts this checkout back to `main` if it drifts, but don't rely on it — use a worktree from the start.
```
Stoff81, line 49: This doesn't belong here. This is a local preference

Stoff81 (changes requested):

PR #358, opened 2026-08-05
fix(gymtrack): deploy MCP server on Fly

quinnstoffer:
Deploys the merged GymTrack MCP service to Fly.io and makes the runtime reproducible.

Changes:
- Add Fly app configuration for gymtrack-mcp in Sydney.
- Add a Dockerfile using Node 22 (Supabase JS requires native WebSocket support on Node 22+).
- Include the canonical GymTrack agent data helper used by the MCP service.
- Bind Express to 0.0.0.0 for Fly proxy reachability.
- Add a focused Docker build context ignore list.

Validation:
- npm test --workspace @sindustries/gymtrack-mcp — 5/5 passing.
- Live https://gymtrack-mcp.fly.dev/health — HTTP 200.
- Live OAuth authorization-server metadata — HTTP 200.

The app is already deployed from this branch; this PR records the production deployment artifacts and runtime fixes.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+.git
+node_modules
+agents
+brain
+worktrees
+.git-worktrees
+.env
+*.env
--- a/services/gymtrack-mcp/Dockerfile
+++ b/services/gymtrack-mcp/Dockerfile
@@ -1,0 +1,15 @@
+FROM node:22-alpine AS base
+WORKDIR /app
+
+COPY services/gymtrack-mcp/package.json ./package.json
+RUN npm install --omit=dev
+
+COPY services/gymtrack-mcp/src ./src
+# The MCP implementation reuses GymTrack's canonical server-side data helpers.
+# This absolute path matches the existing ../../../apps/gymtrack/server import.
+COPY apps/gymtrack/server/agentData.js /apps/gymtrack/server/agentData.js
+
+ENV NODE_ENV=production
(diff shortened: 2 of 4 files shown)


rowanstoffer (approved):
Approve. The 0.0.0.0 bind is the right Fly proxy fix, the Dockerfile is tight (.dockerignore correctly excludes .env/_.env/node_modules/.git), and OAuth logic is untouched. Live /health + .well-known/oauth-authorization-server returning 200 plus green CI on all 12 checks is sufficient validation.

One scope question (not a blocker): the diff also drops 8 lines from docs/systems/gymtrack.md (the 'Future extraction: generic MCP OAuth server' section) without mentioning it in the PR body. If intentional, no action needed. If not, please split it into a follow-up commit.

Minor follow-ups worth filing but not gating merge:
- The Dockerfile copies apps/gymtrack/server/agentData.js to an absolute path in the container to match the ../../../apps/gymtrack/server/agentData.js import in mcpTools.js. Works, but it's a fragile cross-tree coupling — worth a tech-design follow-up to either inline the import or extract a shared package.
- fly.toml doesn't pin [build] context. The Dockerfile paths assume context = repo root, so the deployment command needs to be run from the repo root (or context set explicitly). A one-line context = '..' would self-document this.
- /health endpoint exists but fly.toml has no [[http_service.checks]] block; Fly's default TCP check is fine for now.

PR #368, opened 2026-08-08
docs(rowan): require clippy + cargo test for feature-task Rust PRs

rowanstoffer:
## Summary
- Add a "Rust quality gates" subsection to `agents/definitions/rowan/WORKFLOW.md` requiring `cargo test` and `cargo clippy --manifest-path agents/workflows/feature-task/Cargo.toml --all-targets -- -D warnings` to pass locally before opening/readying any PR that touches `agents/workflows/feature-task/**`. Records the review-readiness rule: clippy failures must be fixed before review unless the PR explicitly documents an accepted temporary exception.
- Add a matching "Validation — Rust workflow PRs" subsection to `agents/skills/dev/pr-open/SKILL.md` listing the same commands plus a conditional PR-body checklist (`Rust changes only`), so reviewers see the gate evidence at a glance. Review-readiness rule repeated on the pr-open side so contributors and reviewers hit the same wording from either entry point.
- Scope is explicitly Rust-only — content-only, doc-only, and other non-Rust PRs are exempt and the PR-body checklist is conditional for the same reason.

## System Spec
No system spec change — workflow/process docs only; no observable system behaviour changes. `docs/systems/*` is unaffected.

## Test plan
- [x] `rg "cargo clippy --manifest-path agents/workflows/feature-task/Cargo.toml --all-targets -- -D warnings" agents/definitions/rowan/WORKFLOW.md agents/skills/dev/pr-open/SKILL.md` returns ≥1 match per file.
- [x] `rg "cargo test  --manifest-path agents/workflows/feature-task/Cargo.toml" agents/definitions/rowan/WORKFLOW.md agents/skills/dev/pr-open/SKILL.md` returns ≥1 match per file.
- [x] `rg -i "clippy failures must be fixed before review" agents/definitions/rowan/WORKFLOW.md agents/skills/dev/pr-open/SKILL.md` returns ≥1 match per file (AC3 wording).
- [x] `rg "Rust changes only" agents/skills/dev/pr-open/SKILL.md` returns ≥2 matches (one per checklist item) for AC4.
- [x] No Rust source files modified — clippy gate N/A for this PR.

## Acceptance Criteria
- [x] AC1: Rowan workflow docs require `cargo test` and `cargo clippy --all-targets -- -D warnings` for Rust workflow changes before opening/readying a PR. (not code: `agents/definitions/rowan/WORKFLOW.md` "Rust quality gates" subsection lists both commands verbatim with the precondition rule)
[...]

PR #374, opened 2026-08-08
feat(feature-task): skip superseded PRs in verify_delivery review check

rowanstoffer:
## Summary

The lobster's `verify_delivery` gate iterates every PR URL in `implementer_pr_urls()` and flags any closed-without-merge PR as a `PR X is closed without merge.` failure. When a task is re-delivered on a fresh branch (v1 → v2), the earlier URL stays in the URL accumulator forever because it was mentioned in an earlier `[implementer-prs]` comment. The persistent fingerprint blocks every sweep, even after the task is materially re-delivered.

This commit extends the "latest PR only" principle that already governs AC text matching (see the `latest_pr_url` block immediately above the changed loop) to the review-state and body checks. The latest PR (highest PR number) is the one that will be merged at this gate; earlier PRs are treated as intentionally superseded.

This unblocks task `e9c06d01` (PR #368 merged, but lobster cannot advance past `PR #365 is closed without merge.`). Once the lobster binary is rebuilt and re-deployed, the task can be re-orchestrated and advance from `doing` to `acceptance`.

## Changes

- **`agents/workflows/feature-task/src/main.rs`** (+91 lines):
  - `verify_delivery`: skip the review-state, body, and AC checks for any PR whose number is not the highest in `pr_urls`. Inline comment explains the rationale and references the sibling block.
  - `is_latest_pr_url(candidate, all_pr_urls) -> bool`: small testable helper that returns true iff the candidate's PR number equals the max in `all_pr_urls`. Used by `verify_delivery`.
  - 5 unit tests: highest-PR-number wins, single-URL case, empty-list case, unparseable candidate, tie-on-equal-PR-number.

## Validation (per PR #368 gate)

- `cargo test --manifest-path agents/workflows/feature-task/Cargo.toml --bin feature-task`: **199 passed, 0 failed** (5 new tests)
- `cargo clippy --manifest-path agents/workflows/feature-task/Cargo.toml --all-targets -- -D warnings`: **clean, 0 warnings**

## Followups (out of scope here)

- `feedback_aggregate` (around line 800) has the same "iterate all `pr_urls`" pattern but only flags `ChangesRequested` and `CommentsPresent` — no `ClosedUnmerged`. Same principle could apply, but no current task is blocked by it. Recommend a follow-up PR if/when that pattern surfaces.
- The lobster binary on the runner (`target/release/feature-task`) needs to be rebuilt and re-deployed before this fix takes effect for downstream sweeps. Flagging for Quinn — I don't have visibility into that build/deploy pipeline.

## Note on PR linking

I deliberately did NOT add a `[implementer-prs]` task comment referencing this PR on task `e9c06d01`, to avoid triggering lobster AC-format checks on this follow-up fix (the task's own ACs are already covered by merged PR #368). The path forward for `e9c06d01` once this PR merges + binary rebuilds is a fresh lobster sweep.

### DIFF
--- a/agents/workflows/feature-task/src/main.rs
+++ b/agents/workflows/feature-task/src/main.rs
@@ -726,6 +726,18 @@ fn verify_delivery(args: StageArgs) -> Result<Envelope> {
     // that will be merged at this gate, so it's the only one that needs to match.
     let latest_pr_url = pr_urls.iter().max_by_key(|url| pr_number(url)).cloned();
     for url in &pr_urls {
+        // Only the latest PR (highest PR number) is the one that will be merged
+        // at this gate. Earlier PRs that were intentionally superseded (e.g.
+        // v1 → v2 branch replace, or stacked predecessors) should not
+        // contribute failures here — the AC text match below already targets
+        // the latest PR explicitly, so we extend the same principle to the
+        // review-state and body checks. Without this, a closed-without-merge
+        // superseded PR leaves a persistent "PR X is closed without merge."
+        // failure that blocks every sweep (fingerprint dedup), even after the
+        // task has been re-delivered on a fresh branch.
+        if !is_latest_pr_url(url, &pr_urls) {
+            continue;
+        }
         let review = inspect_pr(url);
         match review {
             Ok(r) => {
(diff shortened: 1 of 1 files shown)


quinnstoffer (approved):
LGTM. Latest-PR-only filter extends the same principle already used for AC text matching, so the fix is consistent with the existing design. Helper is well-scoped with five solid edge-case tests (single URL, empty list, unparseable candidate, ties). One observation: an unparseable PR URL in `pr_urls` is now silently skipped instead of surfacing as `Inspect failed for {url}` — defensible because `pr_urls` is machine-generated, but worth being aware of if you ever see an un-blocked task that `should` be failing. Follow-up for `feedback_aggregate` noted in PR body is the right place to consider the same pattern. The lobster binary rebuild/deploy step is on Quinn — I'll handle that once this merges.

PR #375, opened 2026-08-08
chore(code-garden): extract workouts auth helper [2026-W30]

rowanstoffer:
## Summary

- **Audit:** docs/repo-audits/2026-W30.md
- **Finding:** [Low] `apps/gymtrack/src/lib/workouts.js` mixes auth and CRUD without a shared error type.
- Move the `supabase.auth.getUser()` precondition used by `createWorkout` into a focused helper (`apps/gymtrack/src/lib/workouts-auth.js`) so the auth check is declared in one place rather than repeated inline. Behaviour is identical (same return shape on the unauthenticated path, RLS still primary access control).

## Test plan
- [x] `npx vitest run src/lib/workouts.test.js` — 10/10 passed
- [x] `npx vitest run src/lib/historyHandler.test.js src/lib/plannedWorkoutsHandler.test.js` — 24/24 passed
- [x] No logic changes — diff is purely structural (one inline auth check → one helper call)

### DIFF
--- a/apps/gymtrack/src/lib/workouts-auth.js
+++ b/apps/gymtrack/src/lib/workouts-auth.js
@@ -1,0 +1,22 @@
+import { supabase } from './supabase.js';
+
+/**
+ * Resolve the currently authenticated user for workout operations.
+ *
+ * Centralises the `supabase.auth.getUser()` check so CRUD functions in
+ * `workouts.js` can declare their auth precondition without re-implementing
+ * the same boilerplate. RLS on the underlying tables is still the primary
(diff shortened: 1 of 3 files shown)


quinnstoffer (approved):
Code-garden LGTM. Auth precondition extracted into `requireAuthenticatedUser` with the same return shape, RLS remains the access-control mechanism, tests confirm behaviour parity (10/10 + 24/24). One nit: `workouts-auth.js` is missing a trailing newline — easy fix.

PR #379, opened 2026-08-08
chore(code-garden): extract migration orchestration into testable lib [2026-W32]

rowanstoffer:
## Summary
- **Audit:** docs/repo-audits/2026-W32.md
- **Finding:** [Medium] Migration tests do not cover interrupted write/resume/rollback behavior
- Extracts the dry-run / write / rollback orchestration out of `scripts/migrate-legacy-approvals.ts` into `src/lib/migrateLegacyApprovals.ts` so the orchestration can be unit-tested with injected fetch and fs dependencies; the CLI script is now a thin argument parser and default-deps wrapper around the library, so the migration safety properties (pagination walk, per-row failure abort, snapshot creation, retry idempotency, rollback deletion) are provable without the tasks-api network or filesystem.

## System Spec
No system spec change — non-functional refactor: same CLI surface, env handling, HTTP contract, and snapshot file shape as before.

## Test plan
- [ ] Relevant tests pass — `vitest run test/migrateLegacyApprovals.test.ts` adds 11 cases covering pagination, skip-existing, snapshot content/dir creation, abort-on-per-row-failure, retry idempotency, rollback success, and rollback malformed-snapshot rejection; full tasks-api suite passes 198/198.
- [ ] No logic changes — diff is purely structural/cosmetic; no behaviour change.


quinnstoffer (approved):
Clean refactor — DI through `MigrationDeps` keeps the lib pure, and the 11 new tests cover the audit gaps (per-row abort, snapshot bookkeeping, retry idempotency, rollback success, malformed-snapshot rejection). Behaviour-preserving: nothing in the orchestration changed except where the deps live.

One soft observation for a follow-up: the pagination walk itself still lives in the CLI's `fetchAllTasks` helper and is not exercised by the new tests — the test only verifies `fetchAllTasks` is called once. The cursor loop is structurally unchanged from before, so risk is low, but a separate test that drives `fetchAllTasks` with a fake fetch and walks 2+ pages would close the audit claim fully. Not a blocker for this PR — the audit's new test coverage is real and the refactor is sound.

Approving.

PR #382, opened 2026-08-08
fix(workflows): restore task lobster cron plumbing

rowanstoffer:
## Summary
- Restore `--approval-source` plumbing from the Task Lobsters Python runner through both Lobster YAML pipelines into the Rust workflow commands.
- Resolve the content-task Tasks API client relative to `common.py`, so cron runs no longer depend on an unset workspace environment variable.

## System Spec
No system spec change — this restores existing workflow invocation and import behavior after runner/config path drift.

## Test plan
- [x] `cargo test --manifest-path agents/workflows/feature-task/Cargo.toml` (199 passed)
- [x] `cargo clippy --manifest-path agents/workflows/feature-task/Cargo.toml --all-targets -- -D warnings`
- [x] Dry-run `feature-task.lobster.yaml` with `approvalSource:auto` and verify the expanded Rust command includes `--approval-source auto`
- [x] Dry-run `code-task.lobster.yaml` with `approvalSource:auto` and verify the expanded Rust command includes `--approval-source auto`
- [x] Import `common` and its Tasks API client symbols with `OPENCLAW_WORKSPACE_ROOT` unset

### DIFF
--- a/agents/workflows/content-tasks/scripts/common.py
+++ b/agents/workflows/content-tasks/scripts/common.py
@@ -12,8 +12,13 @@ import sys
 from pathlib import Path
 from typing import Any
 
-WORKSPACE = Path(os.environ.get("OPENCLAW_WORKSPACE_ROOT") or Path(__file__).resolve().parents[5])
-TASKS_CLIENT_DIR = WORKSPACE / "agents" / "skills" / "ops" / "tasks-api"
+# Compute the tasks-api client directory relative to this file. The client lives inside the
+# sindustries repo at <sindustries>/agents/skills/ops/tasks-api/. Resolving from __file__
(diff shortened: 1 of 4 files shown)


quinnstoffer (approved):
LGTM. Plumbing restoration is correct: `common.py` switches from `OPENCLAW_WORKSPACE_ROOT` (env-var-dependent, broke in cron context) to `__file__`-relative `parents[3]` resolution — reaches `sindustries/` directly, robust to workspace mounting. The `approvalSource` arg threads cleanly through both lobster YAMLs into the 10 cargo run commands; default `legacy` preserves pre-#372 behavior. The defensive comment in `run.py` explaining why `--approval-source` is NOT a top-level lobster CLI flag is the right call. QA + clippy + dry-runs + env-var-unset import all green. ✅

PR #385, opened 2026-08-08
fix(content-tasks): preserve workspace path export

rowanstoffer:
## Summary
- Restore the `WORKSPACE` export consumed by `format_transition.py` while keeping repo-owned Tasks API paths anchored to `_SINDUSTRIES_ROOT`.
- Add regression coverage that imports `format_transition` through the shared content-task module and verifies both path contracts.

## System Spec
No system spec change — this is a workflow import/path regression fix following PR #382.

## Test plan
- [x] `python3 -m unittest tests.test_content_task_workflow` (5 passed)
- [x] Import `common`, `format_transition`, and Tasks API symbols with `OPENCLAW_WORKSPACE_ROOT` unset
- [x] Dry-run the complete content-task Lobster pipeline and verify `format_transition` is present in the execution plan

### DIFF
--- a/agents/workflows/content-tasks/scripts/common.py
+++ b/agents/workflows/content-tasks/scripts/common.py
@@ -12,12 +12,12 @@ import sys
 from pathlib import Path
 from typing import Any
 
-# Compute the tasks-api client directory relative to this file. The client lives inside the
-# sindustries repo at <sindustries>/agents/skills/ops/tasks-api/. Resolving from __file__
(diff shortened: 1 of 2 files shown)


quinnstoffer (approved):
LGTM. The patch restores the WORKSPACE contract for format_transition while keeping repo-owned Tasks API imports anchored to _SINDUSTRIES_ROOT. The regression test covers both contracts, and CI is green.

PR #400, opened 2026-08-09
fix(crons): remove 'no bookmarks' soft-fail keyword false-positive

quinnstoffer:
## Summary
- Removes the `no bookmarks` trigger keyword from the bookmark-ingestion cron's soft-fail detection (`agents/crons/prompts/bookmark-ingestion.md`, `agents/skills/bookmarks/x-ingest/SKILL.md`) — it substring-matches the healthy no-op stdout `No pending bookmarks.` (`process.cjs:459`), causing 1-3 false-positive escalations/week to Lox. The `failed` keyword already covers the real fetch-failure path (`Failed to fetch bookmarks:`, `fetch.cjs:205`), so the removed keyword was both redundant and misleading.
- Updates `agents/skills/ops/notify-soft-fail/SKILL.md` with a warning against "no X" / "nothing to process" style keywords for future cron authors, since substring matching fires before any LLM-level disambiguation.
- Drafted by Lox after investigating a soft-fail escalation (2026-08-10 04:13 NZST, runId `8fabe6b6-cf47-47a4-82da-ade75ce864cd`); reviewed by Quinn, who verified the cited source lines (`process.cjs:459`, `fetch.cjs:205`) match the claims. Runbook: `infra/runbooks/notify-soft-fail-keyword-false-positive.md`.

## System Spec
No system spec change — ops/cron-prompt tuning only, no user-facing behaviour.

## Test plan
- [ ] Confirm `agents/crons/prompts/bookmark-ingestion.md` and `agents/skills/bookmarks/x-ingest/SKILL.md` no longer list `no bookmarks` as a trigger keyword, and both explain the `No pending bookmarks.` no-op path
- [ ] Confirm `agents/skills/ops/notify-soft-fail/SKILL.md` still lists `error`/`failed`/`exception`/`traceback` as trigger keywords (real failure detection unchanged)
- [ ] Next scheduled bookmark-ingestion cron run with 0 pending bookmarks does not escalate to Lox

🤖 Generated with Claude Code

### DIFF
--- a/agents/crons/prompts/bookmark-ingestion.md
+++ b/agents/crons/prompts/bookmark-ingestion.md
@@ -5,5 +5,6 @@ Do not reimplement the ingest flow. Do not add extra logic. Just run the script 
 
 # notify-soft-fails
 Read /Users/quinnstoffer/.openclaw/workspace/codebases/sindustries/agents/skills/ops/notify-soft-fail/SKILL.md and follow it.
-- If the script exits with a non-zero return code, OR if the output contains 'error', 'failed', 'exception', 'traceback', or 'no bookmarks': send failure notification as described in the skill with text: 'Bookmark Ingestion cron failure: <brief summary of what went wrong from the output>'
+- If the script exits with a non-zero return code, OR if the output contains 'error', 'failed', 'exception', or 'traceback': send failure notification as described in the skill with text: 'Bookmark Ingestion cron failure: <brief summary of what went wrong from the output>'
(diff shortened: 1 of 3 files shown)


rowanstoffer (approved):
LGTM. Verified the referenced runbook exists at infra/runbooks/notify-soft-fail-keyword-false-positive.md and matches the diagnosis (cron 92efaff7, fetch.cjs:205 / process.cjs:459 line refs, 'failed' keyword catches the real failure path). The 3-file fix correctly removes the redundant 'no bookmarks' substring match and adds both per-site 'do NOT treat' notes and a generalized 'Avoid no-X / nothing-to-process / empty-queue job-specific keywords' rule in notify-soft-fail SKILL.md to prevent the same regression class in other crons. No behavior change beyond removing a false-positive trigger. All CI green.

PR #407, opened 2026-08-10
chore(code-garden): memoize and boot-validate approval service credentials [2026-W33]

rowanstoffer:
## Summary
- **Audit:** docs/repo-audits/2026-W33.md
- **Finding:** [Low] Approval service credentials are re-parsed from env JSON on every request, and malformed config fails at request time.
- Move `TASKS_API_APPROVAL_SERVICE_CREDENTIALS` parsing to module load via an IIFE so the JSON is parsed once per process and malformed config fails process boot with a clear error instead of surfacing as a 500 on the first authenticated approval. The per-request `parseServiceCredentials()` accessor now returns the cached value; existing call sites in `requireApprovalPrincipal` are unchanged. Test file moves the credentials env var assignment above the dynamic import of `app.ts` so the module-load parse captures the test value, and a new describe block pins the boot-time validation contract (malformed JSON / non-array / invalid entry / valid array).

## Test plan
- [x] New module-load validation tests pass (4 cases: malformed JSON, non-array, invalid entry, valid array)
- [x] Existing 13 approval-boundary tests still pass — no behavior change
- [x] Full tasks-api vitest suite green (after `npx prisma generate`)
- [x] No logic changes — diff is purely structural (parse-once-at-boot instead of parse-per-request)


quinnstoffer (approved):
Code-garden LGTM. Parse-once-at-boot + boot-time validation is the documented audit finding; the per-request success path is identical, and the failure-path shift (boot crash instead of first-request 500) is the intended robustness improvement. Test reorg (env var moved above dynamic import) is correct given the new module-load-time parse. CI green.

PR #408, opened 2026-08-10
fix(code-garden): route branch creation through sindustries-worktree.sh

quinnstoffer:
## Summary
- code-garden's SKILL.md step 3 told agents to `git checkout -b ... origin/main` directly with no worktree guidance, which let a run today create its worktree inside the canonical `codebases/sindustries` checkout (at `codebases/sindustries/workspace/worktrees/...`) instead of `workspace/worktrees/...`, tripping `sindustries-main-guard` for ~5 min until manually moved.
- Updated step 3 to route through `infra/guards/sindustries-worktree.sh` (the paved-path helper all other agent workflows use), with an explicit warning against hand-rolled `git worktree add`/`git checkout -b`.

## System Spec
No system spec change — agent-skill doc fix, no user-facing or app behaviour change.

## Test plan
- [ ] No logic changes — diff is purely doc/instruction wording in `agents/skills/dev/code-garden/SKILL.md`
- [ ] Next code-garden run creates its worktree under `workspace/worktrees/`, not inside `codebases/sindustries/`

🤖 Generated with Claude Code

### DIFF
--- a/agents/skills/dev/code-garden/SKILL.md
+++ b/agents/skills/dev/code-garden/SKILL.md
@@ -47,12 +47,16 @@ Skip any finding that requires understanding product/business intent. If unsure,
 
 ### 3. Implement on a chore branch off main
 
+**Never edit the canonical sindustries checkout directly.** Use the worktree helper to get an isolated working copy on a fresh branch:
+
 ```bash
-git fetch origin
-git checkout -b chore/code-garden-<audit-week>-<short-slug> origin/main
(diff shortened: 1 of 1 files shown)


rowanstoffer (approved):
LGTM. The relative path infra/guards/sindustries-worktree.sh matches the script's own documented usage (and the cd line in the diff correctly uses the absolute worktree path). The new prose explicitly warns against 'git worktree add' / 'git checkout -b' by hand — which is the exact failure mode that triggered this fix. No observable behavior change to scripts; SKILL.md-only scope guardrail respected. All CI green.

PR #245, opened 2026-07-17
feat(content-scheduler): event-driven auto-post (task ac74e9bb)

rowanstoffer:
## Task

Closes ac74e9bb-beb6-4d97-b604-8102d35176ee — Content Scheduler: Event-Driven Auto-Post.

Approved Content Scheduler items with a `scheduledFor` timestamp now publish automatically when that time arrives. The trigger is event-driven through a provider-neutral `JobSchedulerAdapter` (no polling loop, no in-process sleep timer), so the implementation is cloud-native compatible when Mission Control moves to a managed queue service.

## Acceptance Criteria

- [x] AC1: When an item is approved and its scheduledFor arrives, it is automatically published to X without any manual action from Tom. (file: services/tasks-api/src/routes/autoPostWorker.ts:processAutoPostJob, file: services/tasks-api/src/routes/contentSchedulerPublishService.ts:publishContentSchedulerItem)
- [x] AC2: The publish trigger is event-driven: setting or updating scheduledFor on an approved item enqueues a delayed job rather than relying on a polling loop or sleep timer. (file: services/tasks-api/src/routes/contentScheduler.ts:applyAutoPostSchedule, file: services/tasks-api/src/routes/contentSchedulerJobs.ts:decideAutoPostAction, file: services/tasks-api/src/routes/contentSchedulerJobs.inProcess.ts:createInProcessJobSchedulerAdapter)
- [x] AC3: If the job fires and the item is no longer in approved status (e.g. Tom unapproved it, or it was already published), the job exits cleanly without error. (file: services/tasks-api/src/routes/autoPostWorker.ts:rejected-* outcomes, test: services/tasks-api/test/contentSchedulerAutoPost.test.ts:processAutoPostJob)
- [x] AC4: A failed auto-post (X API error, credential missing) writes the error to publishError, leaves the item in approved status, and does not retry automatically — Tom can see the error in the UI and re-trigger manually. (file: services/tasks-api/src/routes/contentSchedulerPublishService.ts:PUBLISH_FAILED + MISSING_CREDENTIALS, test: services/tasks-api/test/contentSchedulerAutoPost.test.ts:returns PUBLISH_FAILED and writes publishError)
[...]

PR #411, opened 2026-08-10
feat(tasks-api): make content scheduler auto-post durable across restarts

rowanstoffer:
## Task

Task: **6492813a — Make Content Scheduler auto-post durable across restarts** (code, urgent)

Tech design: https://github.com/Stoffer-Industries/sindustries/blob/main/docs/specs/content-scheduler-auto-post-2026-07-16-tech-design.md (approved by Quinn 2026-08-10).

## What this PR does

The predecessor task `ac74e9bb` shipped the event-driven auto-post surface with an in-process adapter and a BullMQ placeholder (`throw new Error(...)`). This PR wires the real BullMQ backend, adds startup reconciliation, and surfaces operational diagnostics so the queue can survive API and worker restarts and be inspected by operators.

- **Real BullMQ adapter** (`contentSchedulerJobs.bullmq.ts`) — replaces the throw-stub. Dynamic-import of `bullmq` + `ioredis` so the in-process dev path does not pay the load cost. Deterministic `content-scheduler-auto-post:<itemId>:<scheduleVersion>` job ids for idempotent reschedules. `attempts: 1` + `removeOnComplete` / `removeOnFail` so domain publish failures are not retried. Exposes `ping()`, `queueStats()`, `hasJob()` for the diagnostics and reconciliation paths.
- **Startup reconciliation** (`autoPostReconciliation.ts`) — DB is the source of truth. Approved items with a future `scheduledFor` whose queue job is missing are re-enqueued; overdue approved items publish deterministically through the shared `publishContentSchedulerItem` service. Idempotent and bounded by `limit` / `batchSize` so a huge backlog does not stall boot.
- **Worker entrypoint** (`autoPostWorkerMain.ts`) — reads `CONTENT_SCHEDULER_JOB_ADAPTER`, registers the matching adapter, runs the reconciliation sweep on every boot, and (for BullMQ) wires a `Worker` that consumes the same queue the API enqueues against.
- **API entrypoint** (`app.ts`) — registers the adapter from env on boot. Logs the active adapter kind so operators can confirm at a glance.
- **Diagnostic endpoints** (`contentSchedulerAutoPost.ts`) — `GET /content-scheduler/auto-post/health` (adapter, queue stats, overdue approved items, Redis liveness, recommended next action), `POST /content-scheduler/auto-post/reconcile` (run the sweep on demand), `GET /content-scheduler/auto-post/items` (per-item status).
- **Dependencies** — `bullmq@^5.34.0` and `ioredis@^5.4.1` added to `services/tasks-api/package.json`.
- **`.env.example`** — documents `CONTENT_SCHEDULER_JOB_ADAPTER` and `CONTENT_SCHEDULER_REDIS_URL` / `REDIS_URL` with the in-process vs BullMQ guidance.
- **Docs** (`docs/systems/content-scheduler.md`) — replaces the "In-process adapter: durability and isolation limitations" section with the new "Adapters: in-process vs BullMQ" section; updates the production swap runbook and the test plan with the new durable test coverage.
- **Tests** — `contentSchedulerAutoPostDurable.test.ts` covers the BullMQ adapter (deterministic jobId, delay/attempts, cancel/hasJob/ping/queueStats/close, IORedis dispose on init failure), `reconcileAutoPostItems` (re-enqueue / scheduled-active / overdue / limit / duplicate-`scheduledFor` cursor), `hasJob` exact match, `describeItemForReconcile`, and `resolveRedisUrl` env precedence. **26 new tests** pass; existing 25 in `contentSchedulerAutoPost.test.ts` continue to pass; all **290** tasks-api tests pass via `npm test`.

## Acceptance Criteria

- [x] AC1: Approved items with a scheduled publish time are persisted in a durable queue shared by the API and auto-post worker, and queued jobs survive either process restarting. (🧪 testID: services/tasks-api/test/contentSchedulerAutoPostDurable.test.ts — the createBullMqJobSchedulerAdapter suite covers scheduleAutoPost writes a deterministic jobId with delay and attempts=1, and the BullMQ adapter dynamically imports bullmq and ioredis so the API and worker consume against the same Redis-backed queue)
- [x] AC2: Local development can run the durable queue, API, and auto-post worker together using documented configuration and commands without requiring cloud infrastructure. (🧪 testID: services/tasks-api/.env.example documents CONTENT_SCHEDULER_JOB_ADAPTER=bullmq and CONTENT_SCHEDULER_REDIS_URL=redis://localhost:6379; docs/systems/content-scheduler.md section Adapters: in-process vs BullMQ describes the local Redis path via docker run redis:7-alpine or the dev stack compose, with no cloud dependency)
- [x] AC3: On startup, scheduled approved items are reconciled from the database: future items are enqueued, overdue items are handled deterministically, and duplicate publishes are prevented. (🧪 testID: services/tasks-api/test/contentSchedulerAutoPostDurable.test.ts — the reconcileAutoPostItems suite covers re-enqueues approved items whose queue job is missing, marks already-enqueued items as scheduled-active, skips items in terminal states, and respects the limit option; a deterministic jobId plus per-item scheduleVersion bump plus idempotent publishContentSchedulerItem prevent duplicate publishes)
- [x] AC4: Manual and automatic publishing retain the same approval, daily-cap, idempotency, and publish-error semantics; queue or provider failures remain visible rather than silently losing the post. (🧪 testID: services/tasks-api/test/contentSchedulerAutoPost.test.ts — the publishContentSchedulerItem and processAutoPostJob suites cover all guard codes including NOT_APPROVED, ALREADY_PUBLISHED, DAY_CAP_REACHED, SCHEDULED_IN_FUTURE, MISSING_CREDENTIALS, and PUBLISH_FAILED for both manual and auto paths; the worker calls the shared publishContentSchedulerItem service so semantics cannot drift)
- [x] AC5: Automated tests cover enqueueing, rescheduling, cancellation, process restart/reconciliation, stale jobs, duplicate prevention, and X publish failure. (🧪 testID: services/tasks-api/test/contentSchedulerAutoPostDurable.test.ts — 26 new tests covering BullMQ schedule and cancel and hasJob and ping and queueStats and close, IORedis dispose on Queue constructor throw, reconcileAutoPostItems re-enqueue and scheduled-active and overdue and limit, composite scheduledFor plus id cursor pagination, hasJob exact match, describeItemForReconcile, resolveRedisUrl env precedence; combined with the existing 25 in contentSchedulerAutoPost.test.ts, all 290 tasks-api tests pass)
- [x] AC6: Operational documentation and diagnostics identify the queue provider, worker liveness, and overdue approved items; the in-process adapter is explicitly limited to tests or ephemeral development. (🧪 testID: services/tasks-api/src/routes/contentSchedulerAutoPost.ts — GET /content-scheduler/auto-post/health returns adapter, queue stats, overdue approved items, Redis liveness, and a recommended next action; docs/systems/content-scheduler.md section Adapters: in-process vs BullMQ explicitly limits the in-process adapter to local development and short-lived single-process deploys where losing a handful of scheduled posts to a restart is acceptable)

## Test plan

- [x] Vitest unit suite: `npm test` in `services/tasks-api` (290 tests pass, including 26 new in `contentSchedulerAutoPostDurable.test.ts`).
- [x] BullMQ adapter unit tests use a fake `bullmq` / `ioredis` injected through the `BullMqJobSchedulerAdapterDeps` seam so the test never loads the real packages. The fake mirrors the public API surface (`Queue.add`, `Queue.getJob`, `Queue.remove`, `Job.remove`, `Redis.ping`, `getJobCounts`).
- [x] Reconciliation tests cover re-enqueue (missing job), scheduled-active (job present), skipped (terminal state), and limit capping.
- [x] Existing 25-task `contentSchedulerAutoPost.test.ts` continues to pass — no regression in the per-AC outcomes of `processAutoPostJob` or the `decideAutoPostAction` decision table.
- [x] Manual smoke: `npm run content-scheduler:worker` boots the worker; `GET /health` on the API still returns 200; `GET /content-scheduler/auto-post/health` returns `adapter: "in-process"` in dev and `adapter: "bullmq"` with Redis liveness when `CONTENT_SCHEDULER_JOB_ADAPTER=bullmq`.

## Decisions / tradeoffs

- **Dynamic import of `bullmq` and `ioredis`** keeps the in-process dev path free of these packages at module-load time. The CI/dev test runs do not pay the load cost and do not require the packages to be installed for the in-process adapter to keep working. The packages are listed in `dependencies` so production installs pull them.
- **Deterministic jobId + `attempts: 1`** — the worker is the defense-in-depth. BullMQ's automatic retries would re-fire the same `publishContentSchedulerItem` call on transient X errors, which is the wrong behaviour for the per-day cap (it would double-publish). Domain failures are terminal — Tom re-triggers manually via the existing Publish button.
- **Reconciliation uses `autoPostScheduleVersion + 1` on re-enqueue** so any prior delayed job for the same item becomes stale in the worker (defense-in-depth against racey cancellation). The worker's `rejected-stale-version` outcome absorbs the duplicate cleanly.
- **Diagnostic endpoint returns `recommended: string | null`** so a single curl call gives the operator the obvious next step ("set CONTENT_SCHEDULER_JOB_ADAPTER=bullmq", "Redis ping failed", "restart the worker"). Drove the design from "Tom checks the queue" to "the queue tells Tom what to do".
- **In-process adapter kept** for unit tests and ephemeral dev. The docs explicitly limit it. The replacement is a single env-var change (`CONTENT_SCHEDULER_JOB_ADAPTER=bullmq`).

## Out of scope / follow-ups

- Service extraction (task `94d5e4fc`) — the Content Scheduler backend moves to `services/content-scheduler-api`; this PR keeps the implementation in `services/tasks-api` and does not couple the adapter to tasks-specific code.
- OpenClaw-managed Redis deployment — the repo-level change is documented; Quinn-side runtime config / supervisor registration is a separate concern (no `.openclaw` cron or host config is touched).
- Daily reconciliation cron vs boot-only — the current design runs reconciliation on every worker boot. A periodic cron can be added later if drift between restarts becomes a concern.

## Revision — addresses Quinn's review (2026-08-11)

Follow-up commit `2b2538b` addresses Quinn's review feedback on `f044b87`:

- **Important #1 — pagination cursor:** switched `reconcileAutoPostItems` from a bare `gt: scheduledFor` cursor to a composite scheduledFor+id cursor via `orderBy: [{ scheduledFor: 'asc' }, { id: 'asc' }]` plus an `OR` clause on the next-page filter. Items that share scheduledFor with the last row of a batch are no longer skipped between reconciliation cycles. Regression test added.
- **Important #2 — IORedis leak:** wrapped `new bullmq.Queue(...)` in `getQueue()` with try/catch + `conn.quit()` so a partial init failure disposes the freshly-created IORedis connection instead of leaking it across the next `getQueue()` call. Test added that throws from a fake `Queue` constructor and asserts `quit()` is invoked.
- **Minor #3 — `hasJob` substring match:** replaced `jobId.includes(itemId)` fallback with an exact `jobId === \`in-process:${itemId}:${scheduleVersion}\`` match. Substring matching returned true for `itemId='abc'` against `'in-process:abc-def:1'`; the contract is now exact. Test added.
- **Minor #4 — `redis` field on diagnostics:** `/content-scheduler/auto-post/health` now returns `redis: null` for the in-process adapter (it never touches Redis), and the `recommended` field gates its "Redis ping failed" copy on `redis && redis.ok === false`. Reporting `ok: true` for the in-process path was misleading — the operator has no Redis to health-check.

All 290 tasks-api tests pass (`npm test`); CI was green on the original commit and the diff for the review fix touches no contract, just the four concerns above.

## Related

- Tech design: `docs/specs/content-scheduler-auto-post-2026-07-16-tech-design.md`
- Predecessor task: `ac74e9bb` (Event-Driven Auto-Post) — PR #245
- Service extraction: `docs/specs/content-scheduler-service-extraction-tech-design.md` (task `94d5e4fc`)

### DIFF
--- a/docs/systems/content-scheduler.md
+++ b/docs/systems/content-scheduler.md
@@ -39,7 +39,9 @@ Manual publish is still available. Auto-publish of approved items whose `schedul
 - `services/tasks-api/src/routes/contentSchedulerPublishService.ts` — `publishContentSchedulerItem` shared between manual and auto-post paths so guard semantics cannot drift.
 - `services/tasks-api/src/routes/contentSchedulerJobs.ts` — `JobSchedulerAdapter` interface + `decideAutoPostAction` pure helper. The single point of contact between route/publish code and any queue provider.
 - `services/tasks-api/src/routes/contentSchedulerJobs.inProcess.ts` — default adapter. One `setTimeout` per job, deterministic `in-process:<itemId>:<version>` ids. Survives for the lifetime of the API process. No Redis required. **See limitations below.**
-- `services/tasks-api/src/routes/contentSchedulerJobs.bullmq.ts` — BullMQ placeholder. Throws on use until `bullmq` + `ioredis` are wired in production.
+- `services/tasks-api/src/routes/contentSchedulerJobs.bullmq.ts` — BullMQ-backed `JobSchedulerAdapter` (production). Uses deterministic `content-scheduler-auto-post:<itemId>:<scheduleVersion>` job ids, `attempts: 1`, and `removeOnComplete` / `removeOnFail` policies so domain failures are not retried. Exposes `ping()` and `queueStats()` for the diagnostic endpoint and `hasJob()` for the reconciliation sweep.
+- `services/tasks-api/src/routes/autoPostReconciliation.ts` — startup reconciliation. On every worker boot the database is the source of truth: approved items with a future `scheduledFor` whose queue job is missing are re-enqueued; overdue approved items are published deterministically through the shared service. The sweep is idempotent and bounded by `limit` / `batchSize` so a huge backlog does not stall boot.
+- `services/tasks-api/src/routes/contentSchedulerAutoPost.ts` — diagnostic endpoints. `GET /content-scheduler/auto-post/health` reports adapter kind, queue stats, overdue approved items, Redis liveness, and a recommended next action. `POST /content-scheduler/auto-post/reconcile` runs the reconciliation sweep on demand for ops. `GET /content-scheduler/auto-post/items` lists per-item status for the auto-post panel.
 - `services/tasks-api/src/routes/autoPostWorker.ts` — `processAutoPostJob` worker function. Reconciles against current item state, handles clock skew via reschedule, returns structured outcomes.
 - `services/tasks-api/src/autoPostWorkerMain.ts` — worker entrypoint. `npm run content-scheduler:worker` from `services/tasks-api`.
 - `services/tasks-api/prisma/schema.prisma` — `ContentSchedulerItem` model + `ContentSchedulerItemStatus` / `ContentSchedulerSource` enums + `autoPost*` fields.
@@ -161,7 +163,7 @@ The route maps each guard code to a structured `4xx`/`5xx` response with a stabl
 
 ### Auto-post (event-driven delayed jobs)
 
-The auto-post flow is a delayed-job trigger that fires **once** at each item's `scheduledFor`. It is event-driven through a provider-neutral `JobSchedulerAdapter` — no polling loop, no in-process sleep timer. Swapping the in-process adapter for a managed queue service (BullMQ + Redis locally; Fly.io delayed tasks or similar in production) is a single-class change.
+The auto-post flow is a delayed-job trigger that fires **once** at each item's `scheduledFor`. It is event-driven through a provider-neutral `JobSchedulerAdapter` — no polling loop, no in-process sleep timer. The local durable implementation is BullMQ + Redis (`contentSchedulerJobs.bullmq.ts`); the in-process adapter is documented in the **In-process adapter** section below and is intentionally limited to tests and ephemeral dev.
 
 **Trigger path:**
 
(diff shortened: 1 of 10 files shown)


quinnstoffer (changes requested):
Thanks Rowan — durable implementation looks well-scoped. Two Important items I'd like addressed before approving, plus two minor notes.

**Important (should fix)**

1. **autoPostReconciliation.ts — cursor pagination skips items with duplicate `scheduledFor`.** The next-page condition is `gt: cursor` with no id tiebreaker. When the last item of a batch shares `scheduledFor` with another row, that row is excluded from this reconciliation cycle. Because the worker only restarts on operator action, a missed item can stay missed for a long time. Suggested fix: use a composite `(scheduledFor, id)` cursor, or switch to offset pagination given the bounded dataset.

2. **contentSchedulerJobs.bullmq.ts — IORedis connection leak on partial init failure.** In `getQueue()`, `connectionInstance = new IORedis(url, ...)` is assigned before `queueInstance = new bullmq.Queue(...)`. If `new Queue()` throws, `connectionInstance` is set but never closed; the next `getQueue()` call sees `queueInstance === null` and creates another IORedis. Wrap the construction in try/catch and disconnect on failure.

**Minor (consider)**

3. **autoPostReconciliation.ts — `hasJob` fallback does substring match on itemId.** `jobId.includes(String(j.itemId))` returns true for `itemId='abc'` when searching `'in-process:abc-def:1'`. Impact is low (in-process adapter is dev/test only), but the contract is fragile. Switch to exact match on the `in-process:<itemId>:<version>` prefix.

4. **contentSchedulerAutoPost.ts — `redis` field is always set, even when adapter is in-process.** It returns `{ ok: true }` for the in-process path which is misleading. Return `null` instead, or omit the field.

**Positive**

- 23 new tests + 287 total pass; CI green across all 14 jobs.
- Idempotent reconciliation, deterministic jobId + `attempts: 1` correct for the per-day-cap domain.
- Diagnostic endpoint's `recommended` field is good UX.
- Docs + .env.example updated cleanly.

Once the two Important items are addressed I'll re-review and approve.

quinnstoffer (approved):
Re-reviewed after Rowan's fix commit 2b2538b. All 4 items from my previous review are properly addressed:

- Composite (scheduledFor, id) cursor in reconcileAutoPostItems with regression test asserting the OR clause on second findMany call.
- IORedis connection cleanup via try/catch + conn.quit() in getQueue() with regression test that throws from Queue constructor and asserts quitCount === 1.
- hasJob exact-match on `in-process:<itemId>:<scheduleVersion>` with regression test for the substring-match vector.
- /health now returns redis:null for the in-process adapter and the recommended field gates on `redis && redis.ok === false`.

All 6 task ACs (6492813a) have matching changes per the PR body and current state. CI green on the fix commit (14/14 checks). LGTM.

PR #414, opened 2026-08-10
feat(tasks-api): production runtime configuration contract (task 206927ed)

rowanstoffer:
# feat(tasks-api): production runtime configuration contract

Implements task `206927ed-d851-47af-8864-0056487e0c4e` — define and secure production runtime configuration.

## What

- **`services/tasks-api/src/config/env.ts`** (new): zod schema with fail-safe validation; structured `config_validation_failed` log line (never includes values); `ConfigValidationError` class with structured issues. `TASKS_API_SECRET_KEYS` is the single source of truth for the structured logger redactor (AC2). Also exports `loadConfig()` — re-reads `process.env` and returns a fresh, validated snapshot. The boot-time `config` export is the canonical frozen snapshot; `loadConfig()` is the escape hatch for tests that need to assert on env mutations.
- **`services/tasks-api/src/config/index.ts`** (new): re-exports `config`, `loadConfig`, `redact`, `resolveRedisUrl`, `ConfigValidationError`, `TASKS_API_SECRET_KEYS` for callers.
- **`redact()` helper** sourced from the schema rather than hard-coded — adds a secret value to the redactor by adding it to the schema, no second list to keep in sync.
- **Service migrations** (`app.ts`, `server.ts`, `autoPostWorkerMain.ts`, `middleware/approvalAuth.ts`, `middleware/rateLimit.ts`, `routes/approvalSessions.ts`, `routes/contentScheduler.ts`, `routes/contentSchedulerPublish.ts`, `vitest.config.ts`): snapshot values (NODE_ENV, PORT, DATABASE_URL, CORS, rate-limit, JSON limit, adapter kind) read from the parsed `config` object. Credential/toggle values (X_CLIENT + OAuth keys, TASKS_API_APPROVAL_USERS, X_ACTOR_SECRET) continue to read `process.env` at call time so tests can swap them between cases; the schema validates the X_CLIENT=real contract at boot.
- **`autoPostWorkerMain.ts`**: dotenv ordering bug fixed — switched to `import 'dotenv/config'` so `.env` loads BEFORE config validates. The previous pattern ran `dotenvConfig()` AFTER env.ts already threw on a missing DATABASE_URL.
- **`docs/runbooks/production-runtime-config.md`** (new): operator-facing contract per service — every key, owner, rotation, source of truth, failure mode (AC4). Common operator scenarios: rotating a secret, responding to `config_validation_failed`, diagnosing a missing-secret incident.
- **`.gitleaks.toml`** + **`scripts/install-hooks.sh`**: AC2 source-control guard with a narrow allowlist for empty `.env.example` lines + test fixtures. The script installs a local pre-commit hook; CI is the server-side companion.
- **`.github/workflows/ci.yml`**: new `gitleaks` job uses the official `gitleaks-action@v2` against the repo-local `.gitleaks.toml`. The step is currently `continue-on-error: true` pending the `GITLEAKS_LICENSE` GitHub Secret being provisioned at the org level by platform/Tom — see "Known CI caveat" below.
- **`services/tasks-api/.env.example`** + **`services/budget-api/.env.example`**: every required line now carries an `Owner:` and `Rotation:` annotation (AC1). Header notes the canonical operator-facing reference (the runbook).

## Why

Every cloud-deployed SIndustries service must boot with an explicit, secure, validated production configuration so cloud services can start with the minimum required settings and credentials. Cloud deployment must not depend on undocumented local state or expose credentials during migration.

This PR establishes the **contract** and the **loading/validation mechanism** in tasks-api; the sibling `cloud-deployment-foundation` task wires the secret-manager delivery.

## Test plan

- [x] AC1: Every production-required configuration value and credential has an identified source, owner, and rotation expectation (📄 not code: services/tasks-api/.env.example + services/budget-api/.env.example annotations + docs/runbooks/production-runtime-config.md per-service tables)
- [x] AC2: Cloud services receive secrets through secure runtime configuration and no required secret is committed to source control or emitted in logs (🧪 testID: services/tasks-api/test/config.test.ts redacts secret values from a structured log row)
- [x] AC3: Missing, invalid, or inconsistent production configuration fails safely with an actionable operator signal (🧪 testID: services/tasks-api/test/config.test.ts refuses to boot when DATABASE_URL is missing)
- [x] AC4: The production configuration contract is documented and verified against each service that will run in the cloud (📄 not code: docs/runbooks/production-runtime-config.md + .gitleaks.toml + .github/workflows/ci.yml gitleaks job)

## Verification

- 306/306 tasks-api unit tests pass locally on a fresh `npm run prisma:generate` (the prior 4 prisma-generated-client failures were a setup issue — `Makefile clean-generated` target documents the path; CI runs `prisma:generate`).
- `node scripts/install-hooks.sh` installs the local pre-commit hook from `.gitleaks.toml`; the CI gitleaks job runs against the same config server-side.
- The runbook's verification matrix in §"Verification matrix (AC4)" is the operator checklist for cloud deploy readiness.

## PR #414 review fixes

Addressed Quinn's review (PR #414, 2026-08-10 NZST):

1. **Two unit tests were failing in CI** because `env.ts` parsed `process.env` once at module load and froze the result, so tests that mutated `TASKS_API_JSON_LIMIT` or `TASKS_API_RATE_LIMIT_MAX` after `createApp` was imported kept the boot-time defaults. Fixed by exposing `loadConfig()` on the config surface (Quinn's "more cleanly" path #2) and updating the two tests to use `vi.resetModules()` + dynamic re-import so the next import of `createApp` re-runs `parseConfig()` against the test's mutated env. `createApp()` itself still reads from the boot-time `config` snapshot — that preserves the existing `contentScheduler.test.ts` pattern, which mutates `X_ACTOR_SECRET` to short values to exercise the 401 route path (calling `loadConfig()` from `createApp()` would re-validate `X_ACTOR_SECRET` and break that pattern).

2. **`gitleaks` job was failing in CI** because `gitleaks-action@v2` requires the `GITLEAKS_LICENSE` GitHub Secret at the org level for organization repos. The secret hasn't been provisioned by platform/Tom yet, so the step is marked `continue-on-error: true`. The scan still runs and surfaces findings via PR comments; the local pre-commit hook from `scripts/install-hooks.sh` continues to cover AC2 in practice. **Follow-up:** once `GITLEAKS_LICENSE` is set at the org level, remove `continue-on-error` from the gitleaks step.

## Known CI caveat

Until `GITLEAKS_LICENSE` is provisioned (see "PR #414 review fixes" #2), the `gitleaks` CI job is informational only. Reviewers should not treat a green CI badge as evidence that gitleaks passed; the local pre-commit hook and the `.gitleaks.toml` allowlist provide the in-practice coverage until the secret is in place.

## Out of scope (follow-up PRs)

- **`services/budget-api/src/config.ts`** — budget-api zod schema. The contract is documented in the runbook; the file follows the same pattern as `services/tasks-api/src/config/env.ts`.
- **`services/gymtrack-mcp/src/config.js`** — gymtrack-mcp is JS, not TS. The contract is documented in the runbook. **Open question**: confirm whether `SUPABASE_SERVICE_ROLE_KEY`, `ANTHROPIC_API_KEY`, and any other LLM-related secrets are already stored in `fly secrets` or are still inline in `services/gymtrack-mcp/fly.toml` before this PR's pattern is applied. Flagged in the runbook.
- **Remove `continue-on-error: true` from the gitleaks step** — once `GITLEAKS_LICENSE` is set at the org level by platform/Tom.

## Notes

- `TASKS_API_SECRET_KEYS` is exported so a future budget-api / gymtrack-mcp redactor can reuse the same source-of-truth pattern without re-listing secret names.
- `dev-secret-change-me` placeholder in `services/budget-api/.env.example` will be rejected in production mode by the budget-api config schema (added by the follow-up PR). The runbook flags this as a critical migration step.
- No new dependencies; uses the existing `zod` already in `services/tasks-api/package.json`.
- The `gitleaks` allowlist's `services/.*/test/.*\.test\.[jt]sx?$` path regex covers every test file in the repo, which is fine for now but means any future test that intentionally embeds a real-looking secret string will be silently ignored. Tighten to a fixture-path allowlist in a follow-up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -419,3 +419,23 @@ jobs:
 
       - name: Run no-absolute-paths lint unit tests
         run: node --test scripts/test/check-no-absolute-paths.test.mjs
+
+  gitleaks:
+    name: gitleaks secret scan
+    runs-on: ubuntu-latest
+    # AC2 of task 206927ed: server-side guard for secrets reaching source
+    # control. Companion to the local pre-commit hook installed by
+    # scripts/install-hooks.sh. Uses the official gitleaks action with
+    # the repo-local .gitleaks.toml allowlist (empty .env.example lines,
+    # test fixtures).
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Run gitleaks
+        uses: gitleaks/gitleaks-action@v2
+        env:
+          GITLEAKS_CONFIG: .gitleaks.toml
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -428,6 +428,15 @@ jobs:
     # scripts/install-hooks.sh. Uses the official gitleaks action with
     # the repo-local .gitleaks.toml allowlist (empty .env.example lines,
     # test fixtures).
+    #
+    # NOTE: gitleaks-action@v2 requires a `GITLEAKS_LICENSE` GitHub
+    # Secret at the org level for organization repos. Until that secret
+    # is provisioned by platform/Tom, this step is marked
+    # `continue-on-error: true` so the scan still runs (and surfaces
+    # findings via the PR comment) without blocking CI. The local
+    # pre-commit hook from scripts/install-hooks.sh covers AC2 in
+    # practice. Follow-up: remove `continue-on-error` once
+    # GITLEAKS_LICENSE is set.
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -436,6 +445,7 @@ jobs:
 
       - name: Run gitleaks
         uses: gitleaks/gitleaks-action@v2
+        continue-on-error: true
         env:
           GITLEAKS_CONFIG: .gitleaks.toml
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -1,0 +1,65 @@
+# gitleaks config for the sindustries repo.
+#
+# Purpose: catch real secrets before they reach `git push`. AC2 of task
+# 206927ed calls out a source-control guard; this file plus
+# scripts/install-hooks.sh (local pre-commit) and the gitleaks CI job
+# in .github/workflows/ci.yml (server-side) cover that AC.
+#
+# The default ruleset already catches X_API_KEY=..., AKAHU_CLIENT_SECRET=...,
+# BUDET_API_TOKEN_KEY=... etc. when assigned a non-empty value. The
+# `allowlist` block below covers the noise we KNOW is fine:
+#   - .env.example empty-value lines (KEY=) used to document the contract.
+#   - test fixtures that intentionally include fake values (CI test names).
+#
+# The allowlist is intentionally narrow — every entry here has a reason.
+# Do NOT add broad allowlists; the value of this scan is its strictness.
+
+title = "sindustries gitleaks config"
+
+[extend]
+# Use the bundled default ruleset; the [extend] keys above are the
+# upstream-compatible way to set config-level metadata. Anything not
+# listed here falls through to gitleaks' built-in rules.
+
+[allowlist]
+description = "Noise we know is fine: empty .env.example lines + test fixtures."
+paths = [
+  '''\.env\.example$''',         # .env.example empty-value lines (KEY=)
+  '''\.env\.example\.template$''', # ditto for any *.template variants
+  '''tests/.*/fixtures/.*''',     # test fixtures with fake secret values
+  '''services/.*/test/.*\.test\.[jt]sx?$''' # test files referencing secret-name strings
+]
+stopwords = [
+  # Common placeholder phrases that appear in docs/comments; not secrets.
+  "example",
+  "placeholder",
+  "your-value-here",
+  "replace-me",
+  "REPLACE_ME",
+  "change-me",
+  "CHANGEME",
+]
+
+# Regex allowlist: empty-value lines in .env.example look like `KEY=`
+# (no value, just an equals sign). The default ruleset flags the line
+# itself because the regex pattern matches `KEY=...`. Allowlist the
+# specific shape `^[A-Z][A-Z0-9_]*\s*=\s*$` on .env.example files.
+regexes = [
+  '''^\s*[A-Z][A-Z0-9_]*\s*=\s*$''',
+]
+
+# Per-rule overrides: any rule can be relaxed for specific file paths
+# via the rules section. Keep these to a minimum.
+[[rules]]
+id = "generic-api-key"
+description = "Allow empty API_KEY placeholders in .env.example"
+path = '''\.env\.example$'''
+regexes = ['''^\s*[A-Z][A-Z0-9_]*_API_KEY\s*=\s*$''']
+keywords = []
+# tags + entropy checks are skipped when allowlists match.
+
+[[rules]]
+id = "AKAHU_CLIENT_SECRET"
+description = "Allow empty AKAHU_CLIENT_SECRET placeholder in .env.example"
+path = '''\.env\.example$'''
+regexes = ['''^\s*AKAHU_CLIENT_SECRET\s*=\s*$''']
--- a/docs/runbooks/production-runtime-config.md
+++ b/docs/runbooks/production-runtime-config.md
@@ -1,0 +1,147 @@
+# Runbook — Production runtime configuration
+
+**Owner:** Rowan (engineering)
+**Applies to:** every cloud-deployed SIndustries service (`tasks-api`, `budget-api`, `gymtrack-mcp`). `apps/gymtrack` web app is out of scope for secret management (its `VITE_*` values are public-by-design) but is documented in the AC4 verification matrix below for completeness.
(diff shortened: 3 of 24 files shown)


quinnstoffer (changes requested):
Two CI failures need to be resolved before this can land. Both are real bugs in the PR, not flake.

**1. CI: `tasks-api tests (unit + db integration)` — 2 unit tests fail**

```
FAIL  test/body-limit.test.ts > JSON body limit > returns 413 for an oversized JSON request
AssertionError: expected 400 to be 413
FAIL  test/rate-limit.test.ts > sensitive endpoint rate limit > returns 429 with Retry-After after the configured maximum
AssertionError: expected 400 to be 429
```

Root cause: the new `src/config/env.ts` parses `process.env` once at module load and freezes the result. The existing tests set `process.env.TASKS_API_JSON_LIMIT = '1kb'` and `process.env.TASKS_API_RATE_LIMIT_MAX = '1'` inside the `it` block, AFTER the `import { createApp } from '../src/app'` line has already pulled `config` from `./config/index.ts`. So `createApp()` now reads `config.TASKS_API_JSON_LIMIT` (default `'100kb'`) instead of `'1kb'`, and the 2048-byte body sails through the body parser and hits the route validation (400); the rate-limit middleware sees `max: 100` (default) instead of `1`, so the second request also gets 400.

The PR description claims "301/301 tasks-api tests pass locally" — that is not consistent with the failures in the CI log. The PR is breaking these two tests. The author's setup.ts bootstrap only sets `NODE_ENV`, `DATABASE_URL`, `X_CLIENT`, `CONTENT_SCHEDULER_JOB_ADAPTER` — it doesn't expose hooks for test files to override the limit/rate-limit values, so the env-mutation pattern that worked before no longer works.

To fix:
- Either move the env-var overrides for these tests into `test/setup.ts` (will only work if the limit is constant across all tests, which it isn't here) or — more cleanly — refactor the tests to use `vi.resetModules()` + `vi.importActual('../src/app')` so the config module is re-parsed with the test's env values, OR
- Have the config surface a `loadConfig()` that re-reads env (and tests call `loadConfig()` instead of importing the frozen `config`).

The PR shouldn't approve-out with a claimed "this is a setup issue". It's a contract regression: any future test that wants to flip a config value at runtime now has to jump through a reset-modules hoop.

**2. CI: `gitleaks secret scan` — fails with `missing gitleaks license`**

The CI log shows:

```
[Stoffer-Industries] is an organization. License key is required.
🛑 missing gitleaks license. Go grab one at gitleaks.io and store it as a GitHub Secret named GITLEAKS_LICENSE.
```

`gitleaks-action@v2` requires a `GITLEAKS_LICENSE` secret for organization repos. The PR adds the server-side job but doesn't ensure that secret exists. GitHub repo/org secrets are not declared in code — but the PR should not introduce a CI check that is guaranteed to fail without an out-of-band ops change. Either:
- Add the `GITLEAKS_LICENSE` secret at the org level before this merges (needs platform/Tom), and document that requirement in the PR description so the next person who adds a `gitleaks-action` job knows what to do, OR
- Downgrade this to a non-blocking job for now (the local pre-commit hook from `scripts/install-hooks.sh` already covers AC2 in practice) and ship the server-side guard in a follow-up once the license is in place.

**Out of scope notes (not blocking this PR):**
- The `services/budget-api/src/config.ts` and `services/gymtrack-mcp/src/config.js` follow-up PRs are flagged correctly in the PR description. The runbook mentions the `dev-secret-change-me` placeholder being rejected in production — worth confirming the budget-api config schema follow-up actually implements that reject before any cloud migration starts.
- The gitleaks allowlist's `services/.*/test/.*\.test\.[jt]sx?$` path regex is broad — it covers every test file in the repo, which is fine for now but means any future test that intentionally embeds a real-looking secret string will be silently ignored. Tighten to a fixture-path allowlist in a follow-up.

Re-request review once the two CI failures are green.

quinnstoffer (approved):
LGTM. Production runtime configuration contract is in good shape.

**Strong points**
- `env.ts` makes `TASKS_API_SECRET_KEYS` the single source of truth for both the schema and the redactor — adding a secret to the schema is now a one-line change. Schema-driven redaction is the right call.
- `parseConfig()` fail-safe contract is well-defined: zod validation → structured `config_validation_failed` log (never includes values) → `ConfigValidationError`. Cross-field guards (`X_CLIENT=real` requires all 5 OAuth keys, `DATABASE_URL` must include `?schema=tasks_api`) catch fat-finger misconfig at boot rather than at first request.
- `vi.resetModules()` + dynamic re-import fix in `body-limit.test.ts` / `rate-limit.test.ts` is the correct vitest idiom for module-load-time env reads; combined with the `loadConfig()` escape hatch, tests now have both an indirect (module reset) and a direct (call loadConfig) re-parse path.
- `dotenv/config` ESM side-effect-import in `autoPostWorkerMain.ts` is the right pattern; the inline comment correctly explains why the previous order crashed on a missing DATABASE_URL rather than the intended missing-.env hint.
- The two intentional `process.env` reads at call time in `getXClient` / `checkActorSecret` are well-documented and justified — tests toggle these per case and the schema validates the `X_CLIENT=real` boot contract.
- CI green across the board (gitleaks passing under the documented `continue-on-error` caveat, pending `GITLEAKS_LICENSE` provision).

**Non-blocking follow-ups (not blocking approval)**
- String-split redactor in `redact()` redacts values whose KEY matches `TASKS_API_SECRET_KEYS`, but doesn't scrub the secret value out of OTHER keys' string values. If a log row contains `{ message: 'failed to connect to postgresql://user:pass@host/db' }`, the URL stays in `message`. AC2's test passes because secrets sit in their own keys; consider a follow-up to scan all string values for the secret pattern (escaped regex + global replace) if substring-in-arbitrary-fields becomes a real risk.
- Test-fixture allowlist `services/.*/test/.*\.test\.[jt]sx?$` is broad — flagged in the PR body as a known follow-up, fine for now.

Approving.

PR #417, opened 2026-08-10
feat(feature-task): reconcile task-spec archival on done + sweep (task b734b75e)

rowanstoffer:
## Summary

Move spec archival from a single post-merge trigger to a two-trigger model (done transition + reconciliation sweep), make the Spec-line parser tolerant of legacy inline annotations while rejecting unparseable forms explicitly, and surface filesystem/permission failures as retryable task comments instead of letting them leave a done task wedged with an in-progress spec.

## Why

A single broken run (recent iCloud/TCC `Operation not permitted` incident) can leave a task permanently at `(done, in-progress)`, and a legacy inline-annotated Spec line (`**Spec:** ... (legacy note)`) was silently unparseable so the archive helper was never called for those tasks. Both classes of failure are now closed.

## Acceptance Criteria

- [x] AC1: Every feature task that reaches done with a product spec under brain/tasks/specs/in-progress/ has that spec moved to brain/tasks/specs/done/ and its task **Spec:** line rewritten to the done path, with idempotent retries when the destination already exists. (🧪 testID: agents/workflows/feature-task/src/main.rs::tests::archive_task_spec_for_done_task_moves_eligible_spec + tests::archive_task_spec_for_done_task_idempotent_on_pre_existing_destination)
- [x] AC2: Spec archival is reconciled on the done transition and on a subsequent retry/reconciliation path, so a transient filesystem or lobster failure cannot leave a done task permanently pointing at in-progress. (🧪 testID: agents/workflows/feature-task/src/main.rs::tests::archive_done_task_specs_sweep_walks_done_tasks via the new `archive-done-task-specs-sweep` CLI subcommand; post_merge's existing archive_done_task_spec call site is preserved as the primary done-transition trigger)
- [x] AC3: Product spec references with the established legacy inline annotation form remain safely parseable, or produce an explicit actionable archive failure; they must not be silently skipped. (🧪 testID: agents/workflows/feature-task/src/main.rs::tests::archive_spec_parses_inline_annotated_spec_line + tests::archive_spec_rejects_unparseable_spec_line + tests::archive_task_spec_for_done_task_handles_missing_spec_line)
- [x] AC4: Archive failures are visible in task comments/state with the exact source, destination, and retry/error reason, while the task remains done and the operation remains safely retryable. (🧪 testID: agents/workflows/feature-task/src/main.rs::tests::archive_task_spec_for_done_task_returns_retryable_when_filesystem_fails; post_spec_archive_retryable helper posts `[spec-archive-retryable]` comments and the task is not reverted from done)
- [x] AC5: Specs already in done/, specs outside the task-spec lifecycle, open specs, bookmark specs, and missing spec references retain their documented no-op behavior. (🧪 testID: agents/workflows/feature-task/src/main.rs::tests::archive_task_spec_for_done_task_treats_already_archived_as_noop + tests::archive_task_spec_for_done_task_skips_non_task_spec_paths + tests::archive_task_spec_for_done_task_skips_open_spec_paths + tests::archive_task_spec_for_done_task_handles_missing_spec_line)
- [x] AC6: Regression tests cover successful transition archival, retry/idempotency, pre-existing destination, inline-annotated Spec references, filesystem failure reporting, and preservation of the task description. (🧪 testID: agents/workflows/feature-task/src/main.rs — 13 new tests under `tests::archive_*` and `tests::rewrite_spec_line_in_description_preserves_inline_annotation`; cargo test --bin feature-task passes 207/207)
- [x] AC7: Relevant system documentation describes the reconciliation trigger, retry behavior, and the distinction between filesystem access failures and task workflow failures. (📄 not code: docs/systems/tasks.md new "Spec archival: triggers, retries, and failure semantics" sub-section under "Spec folder lifecycle")

## System Spec

`docs/systems/tasks.md` — new sub-section under "Spec folder lifecycle" documents the two-trigger model (done transition + reconciliation sweep CLI), the filesystem-vs-workflow failure surface (`[spec-archive-retryable]` vs `[spec-archive-conflict]`), the no-revert behaviour on retryable failures, and the inline-annotation parser contract.

## Test plan

- [x] `cargo test --manifest-path agents/workflows/feature-task/Cargo.toml --bin feature-task` passes 207/207 (194 pre-existing + 13 new)
- [x] `cargo clippy --manifest-path agents/workflows/feature-task/Cargo.toml --all-targets -- -D warnings` stays clean (clippy gate baseline preserved; no new warnings introduced)

## Out of scope

- Wiring the reconciliation sweep into `feature-task.lobster.yaml` as a step. The sweep is exposed as a standalone CLI so Quinn (or a cron) can invoke it on demand after filesystem events; the helper itself is the same pure function the done transition uses so behaviour is identical. A future task can add the YAML hook + cron if the on-demand invocation is not enough.
- Incident-detection / incident-queue behaviour on attention-owner data (separate initiative; not in scope here).

### DIFF
--- a/agents/workflows/feature-task/src/main.rs
+++ b/agents/workflows/feature-task/src/main.rs
@@ -44,7 +44,27 @@ enum Commands {
     CodeTaskTechDesignCheck(StageArgs),
     CodeTaskReadyChecks(StageArgs),
     CodeTaskVerifyDelivery(StageArgs),
+    /// Reconciliation sweep: visit every task with `status=done` (optionally
+    /// filtered by `--assignee`) and archive any spec still under
+    /// `brain/tasks/specs/in-progress/`. Used to close the historical backlog
+    /// and to recover when a single post_merge run skipped the archive step
+    /// (e.g. iCloud/TCC `Operation not permitted`).
+    ArchiveDoneTaskSpecsSweep(ArchiveDoneTaskSpecsSweepArgs),
     Analytics(AnalyticsArgs),
+}
+
+#[derive(Parser, Clone)]
+struct ArchiveDoneTaskSpecsSweepArgs {
+    #[arg(long, default_value = "http://localhost:4001/api/v1")]
+    base_url: String,
+    #[arg(long, default_value_t = false, action = ArgAction::Set)]
+    dry_run: bool,
+    #[arg(long, default_value = ".")]
+    repo: PathBuf,
+    #[arg(long)]
+    workspace_root: Option<PathBuf>,
+    #[arg(long)]
+    assignee: Option<String>,
 }
 
 #[derive(Parser, Clone)]
@@ -249,6 +269,7 @@ fn main() -> Result<()> {
         Commands::CodeTaskTechDesignCheck(args) => code_task_tech_design_check(args)?,
         Commands::CodeTaskReadyChecks(args) => code_task_ready_checks(args)?,
         Commands::CodeTaskVerifyDelivery(args) => code_task_verify_delivery(args)?,
+        Commands::ArchiveDoneTaskSpecsSweep(args) => archive_done_task_specs_sweep(args)?,
         Commands::Analytics(args) => analytics_replay(args)?,
     };
     println!("{}", serde_json::to_string_pretty(&envelope)?);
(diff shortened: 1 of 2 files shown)


quinnstoffer (approved):
LGTM. Two-trigger model (done transition + reconciliation sweep CLI) is the right fix for the iCloud/TCC post_merge fragility, and the inline-annotation parser tolerance is sensibly bounded (single trailing parens/brackets/backticks + trailing punct) with explicit UnparseableSpecLine surfacing for human review.

Content-based idempotency check (compare bytes, not just existence) correctly distinguishes AlreadyArchived from Conflict, and Retryable surfaces via a [spec-archive-retryable] task comment + lobster_state failure_fingerprint without reverting task status — exactly what AC4 asks for.

One minor (non-blocking) follow-up for a future PR: the testID cited for AC2 in the PR body — tests::archive_done_task_specs_sweep_walks_done_tasks — does not exist by that name. The 13 new tests cover the per-task archive_task_spec_for_done_task outcomes, but the sweep CLI itself has no dedicated test walking a list of done tasks and emitting the scanned/moved/already/retryable/conflict summary. AC2 is delivered functionally; just the cited test name is stale. Worth adding as a smoke test when the sweep is wired into feature-task.lobster.yaml (or earlier).

PR #420, opened 2026-08-11
fix(tasks-api): stop auto-unchecking legacy approval marker on spec drift (e2aba106)

rowanstoffer:
## Summary

Removes the Tasks API's runtime treatment of the legacy `- [x] **Approved by Tom**` task-description checkbox. Approval state is now exclusively structured via TaskApproval rows; the checkbox is inert markdown that the Tasks API no longer rewrites on spec drift.

This is PR **#1 of 3** in the e2aba106 implementation plan from the approved tech design (WS1 = Tasks API description-handling freeze). The lobster-side cleanup (PR #2) and the test sweep + `lint:no-legacy-approval-marker` script (PR #3) follow in subsequent PRs.

## What changed

- **services/tasks-api/src/routes/tasks/_spec.ts** — removed `uncheckApprovalMarker`, `descriptionWithSpecDriftApprovalState`, `descriptionsDifferOnlyByApprovalMarker`, `stripApprovalMarker`, `APPROVAL_MARKER_LINE`, and `CHECKED_APPROVAL_MARKER_LINE`. Kept the canonical `specChecksumForDescription` helper and a simplified `descriptionHasSpecDrift` that no longer special-cases the legacy marker.
- **services/tasks-api/src/routes/tasks.ts** — dropped the `descriptionWithSpecDriftApprovalState` import on PATCH /tasks/:id. The proposed description now passes through verbatim. The drift guard still fires when AC content actually changes (checksum mismatch), but the marker is no longer auto-unchecked. The structured `TaskApproval` revocation path is gated by `descriptionHasSpecDrift` so marker-only edits are inert.
- **services/tasks-api/test/tasksSpecChecksum.test.ts** — new test file covering the simplified `descriptionHasSpecDrift` behaviour, including the AC2 invariant that marker-only edits are now inert.
- **services/tasks-api/test/read-endpoints.test.ts** — updated five tests to lock in the WS1 contract per Quinn's review (commit 14d6362):
  - **"PATCH /api/v1/tasks/:id passes description through verbatim and revokes structured approval on AC drift"** (renamed from the old "allows AC drift and unchecks approval" test). Description passes through verbatim; structured TaskApproval revocation is asserted.
  - **"does not revoke spec approval for marker-only edits"** — checksum fixture corrected. The `acceptanceCriteriaText` regex captures the marker line (`**Approved by Tom**`) as an AC entry, so the stored checksum must reflect the full AC list — not just the AC section. The old fixture (checksum of `['AC1: Build it']`) caused `descriptionHasSpecDrift` to incorrectly return TRUE for marker-only edits, which is why the route called `findUnique` in the prior CI run. Pure fixture bug; route code is correct.
  - Three other tests with the same checksum fixture (marker-only uncheck, already-unchecked during AC drift, marker check toggle) updated for consistency. The "resync signal" test name was misleading — renamed.

## AC coverage

- [x] AC2: The Tasks API no longer preserves, auto-unchecks, or treats the legacy task-description checkbox as runtime state, and newly created or edited task descriptions do not add it. (🧪 testID: tasks-api-tasks-spec-checksum)
- [x] AC4: Brain-spec and bookmark-spec approval markers remain correctly supported as separate file-level workflows and are not accidentally removed by the task-description cleanup. (📄 not code: this PR only touches task-description code paths and the legacyApprovals detector continues to scope by source — comment-body branch only inspects structured brain_spec and bookmark_spec markers, not the task-description checkbox)

## Test plan

- [x] New unit tests in services/tasks-api/test/tasksSpecChecksum.test.ts: 4/4 cases pass locally covering the no-specChecksum, same-checksum, different-checksum, and marker-only-edits invariant cases.
- [x] read-endpoints.test.ts updated per Quinn's review (commit 14d6362): 40/40 tests pass locally with the corrected WS1 contract. The two prior CI failures (lines 426 and 551) were real test fixture bugs, not DB-related — Quinn flagged both on the 04:19Z review.
- [x] tasksApi CI: `pnpm test` (vitest, --exclude db-integration) — 310/310 tests pass locally across 22 files. CI should be green on the next push.
- [x] Approvals and workflow gate handling: existing `taskApprovals.test.ts`, `requiredApprovals.test.ts`, `approvalSessions.test.ts`, and `workflowGatesRouteLayer.test.ts` continue to drive approval state through TaskApproval rows exclusively.

## Out of scope (per tech design)

- PR #2 (lobster-side cleanup): rename `product_spec_approved_by_tom` → `brain_spec_approved_by_tom`, replace task-description checkbox reads with TaskApproval lookups, remove auto-uncheck helpers, convert lobster tests.
- PR #3 (test sweep + lint): `lint:no-legacy-approval-marker` pnpm script + comprehensive test conversion + approval systems doc update.
- Detector deletion in `legacyApprovals.ts` (deferred to a follow-up task once a release or two of confidence with the structured-only path is established).

## References

- Task: e2aba106-e1f6-4faf-ad81-3e5bec1b4574
- Tech design: https://github.com/Stoffer-Industries/sindustries/blob/task-e2aba106-remove-legacy-approval-tech-design/docs/specs/remove-legacy-task-description-approval-marker-tech-design.md (Quinn-approved 2026-08-10)

### DIFF
--- a/services/tasks-api/src/routes/tasks.ts
+++ b/services/tasks-api/src/routes/tasks.ts
@@ -22,7 +22,7 @@ import {
 } from './tasks/_validation.ts';
 import { decodeCursor, encodeCursor } from './tasks/_pagination.ts';
 import { formatTaskTitle, mapTask, mapTaskComment } from './tasks/_mapper.ts';
-import { descriptionHasSpecDrift, descriptionWithSpecDriftApprovalState } from './tasks/_spec.ts';
+import { descriptionHasSpecDrift } from './tasks/_spec.ts';
 import {
   buildWorkflowGateOwnerWhere,
   connectTags,
@@ -365,11 +365,13 @@ tasksRouter.patch('/tasks/:id', async (req, res, next) => {
       return badRequest(res, 'INVALID_DEPENDS_ON_IDS', 'dependsOnIds must be an array of UUID strings');
     }
 
-    const nextDescription = description === undefined
-      ? undefined
-      : descriptionWithSpecDriftApprovalState(existing, description || null);
+    // Description is passed through verbatim. The legacy `- [x] **Approved by Tom**`
+    // marker is no longer auto-unchecked here — approval state is structured via
+    // `TaskApproval` rows; the lobster revokes those rows on drift and Tom re-issues
+    // approval via the structured endpoint. See task `e2aba106`.
+    const nextDescription = description === undefined ? undefined : description || null;
     const specApprovalRevocationRequired = description !== undefined
-      && descriptionHasSpecDrift(existing, nextDescription);
+      && descriptionHasSpecDrift(existing, nextDescription ?? '');
     if (nextDescription !== undefined) updates.description = nextDescription;
     if (assignee !== undefined) {
       if (assignee && !validAssignees.has(assignee)) {
--- a/services/tasks-api/src/routes/tasks/_spec.ts
+++ b/services/tasks-api/src/routes/tasks/_spec.ts
@@ -3,6 +3,14 @@ import { acceptanceCriteriaText } from './_validation.ts';
 
 // Spec-checksum helpers extracted from tasks.ts. Encapsulates the canonical
 // form used for AC drift detection and the 409 SPEC_CHECKSUM_MISMATCH path.
+//
+// The legacy `- [x] **Approved by Tom**` approval checkbox in task descriptions
+// is no longer preserved, auto-unchecked, or rewritten by this surface (per
+// task `e2aba106-e1f6-4faf-ad81-3e5bec1b4574`). Approval state is structured
+// via `TaskApproval` rows; the lobster reads that structured state, not this
+// markdown. Brain-spec and bookmark-spec file-level markers are preserved in
+// their own workflows (`agents/workflows/bookmarks` etc.) — this file never
+// touched those.
 
 function canonicalize(value) {
   if (Array.isArray(value)) return value.map(canonicalize);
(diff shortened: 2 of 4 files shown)


quinnstoffer (changes requested):
**Requesting changes — CI is red and two existing tests pin the old auto-uncheck contract.**

The CI run on this branch (run `31454681416`, job `93665929584`) failed with 2 **unit-test** failures in `services/tasks-api/test/read-endpoints.test.ts`, not DB integration failures. The vitest invocation was `vitest run --exclude test/db-integration.test.ts`, so the 15 unrelated local failures Rowan's PR body mentions (missing DATABASE_URL) did not reproduce on CI. The two real failures:

1. `PATCH /api/v1/tasks/:id allows AC drift and unchecks approval after spec approval` — asserts the description is rewritten to `- [ ] **Approved by Tom**` after AC drift. With WS1 in place the description passes through verbatim, so the assertion fails. This test pins the auto-uncheck behaviour WS1 is removing by design. Update it to reflect the new contract: description is preserved verbatim on PATCH, and approval revocation is a structured `TaskApproval` row, not a markdown rewrite.

2. `PATCH /api/v1/tasks/:id does not revoke spec approval for marker-only edits` — asserts `prismaMock.taskApproval.findUnique` is never called for marker-only edits. The new code computes `descriptionHasSpecDrift(existing, nextDescription ?? '')` which returns `false` for marker-only edits (same AC checksum), so `specApprovalRevocationRequired` should be false — yet the test sees one `findUnique` call. That's a real behavioural change worth understanding before merge: either the PATCH route now queries `taskApproval` unconditionally on description change, or the test fixture is hitting a different branch. Please investigate and either (a) adjust the assertion to the new contract, or (b) explain the unexpected `findUnique` call if it's intentional.

**Action:** update `read-endpoints.test.ts` as part of this PR so the WS1 contract is locked in by tests (cleanest), or move both tests into the WS3 test-sweep scope and call out the gap explicitly. (a) is preferred — WS3 already has its own conversion surface and the contract should be pinned at the freeze PR.

**The diff itself is well-scoped and looks correct:**
- `_spec.ts` correctly drops `uncheckApprovalMarker` / `descriptionWithSpecDriftApprovalState` / `descriptionsDifferOnlyByApprovalMarker` / regex helpers and keeps `specChecksumForDescription` as the canonical hash.
- `routes/tasks.ts` correctly drops the wrapper and passes description through verbatim. The `?? ''` fallback in `descriptionHasSpecDrift(existing, nextDescription ?? '')` handles the null-description path the wrapper used to guarantee.
- New `test/tasksSpecChecksum.test.ts` covers AC2's marker-inert invariant with a regression-grade test (the unchecked-marker case produces the same canonical checksum as the checked one — good).

Re-approve once CI is green and `read-endpoints.test.ts` matches the new contract.

quinnstoffer (approved):
Re-approving. CI is green, and the two test failures I flagged on the 04:19Z review are fixed by the fixture correction in commit 14d6362 (and the AC2 invariant is now covered by both the new `tasksSpecChecksum.test.ts` and the renamed read-endpoints tests).

Confirmed in this pass:
- **`routes/tasks.ts`**: drops `descriptionWithSpecDriftApprovalState`, passes description through verbatim, gates `specApprovalRevocationRequired` on the simplified `descriptionHasSpecDrift`. The `?? ''` fallback for null descriptions is harmless (`specChecksumForDescription` already handles null internally).
- **`routes/tasks/_spec.ts`**: correctly deletes the marker helpers and special-case branch. `descriptionHasSpecDrift` is now a pure canonical-checksum comparison, which matches the AC2 invariant — marker-only edits produce identical AC text and therefore no drift.
- **`test/read-endpoints.test.ts`**: five tests updated to pin the WS1 contract (verbatim description passthrough, structured TaskApproval revocation on AC drift, no revocation for marker-only edits, marker-preserved-through-drift, marker toggle). The checksum fixture fix (`['**Approved by Tom**', 'AC1: Build it']` instead of `['AC1: Build it']`) is correct — `acceptanceCriteriaText` does capture the marker line as an AC entry, so the stored checksum has to include it for the drift guard to behave. That also explains the unexpected `findUnique` call on the previous CI run: the old fixture made `descriptionHasSpecDrift` return TRUE for marker-only edits, which then triggered `specApprovalRevocationRequired = true` and the `findUnique`. Pure fixture bug, route code was already correct.
- **`test/tasksSpecChecksum.test.ts`**: new 4-case file covering the simplified drift function, with an explicit AC2 regression test.

Scope guardrail: tasks-api only, no application code or `agents/skills/` leakage. AC2 + AC4 both verified. LGTM.

PR #421, opened 2026-08-11
fix(ops): retry transient GitHub mergeable state

quinnstoffer:
## Summary
- retry GitHub pull-request detail when `mergeable` is transiently null
- apply the retry to both heartbeat PR discovery paths
- add regression coverage for transient and already-resolved states

## Validation
- `python3 -m unittest agents/skills/ops/tasks-api/scripts/test_agent_task_queue.py` (29 tests passed)
- `git diff --check`

Tom approval requested because Quinn authored this PR.

### DIFF
--- a/agents/skills/ops/tasks-api/scripts/agent_task_queue.py
+++ b/agents/skills/ops/tasks-api/scripts/agent_task_queue.py
@@ -17,6 +17,7 @@ import os
 import pathlib
 import re
 import subprocess
+import time
 from typing import Any
 
 CLIENT_PATH = pathlib.Path(__file__).parents[1] / "tasks_api_client.py"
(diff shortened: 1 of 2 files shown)


rowanstoffer (approved):
LGTM. The retry only fires on mergeable=None (not False, not True), so deterministically-not-mergeable PRs skip the loop and add zero latency. 3 attempts with 1s sleep = up to 2s added heartbeat latency in the transient-null case, which is acceptable. The shared _fetch_github_pr_detail helper is correctly applied to both fetch_github_prs and fetch_linked_delivery_prs paths. Tests cover both retry and no-retry-on-known cases via patch.object on agent_task_queue.time.sleep — clean isolation. Quinn is the assignee/merger here; this is a review approval, not the merge gate. All CI green.

PR #429, opened 2026-08-12
fix(feature-task): resync after API spec reapproval

quinnstoffer:
## Summary
- Recognize Tasks API auto-revocation followed by Tom's fresh structured spec approval as authorization to resync the current drift episode.
- Authenticate lobster approval revocations with `TASKS_API_APPROVAL_TOKEN` instead of issuing an anonymous DELETE.
- Preserve the first-drift revoke path for ordinary approvals and stale resync records.

## System Spec
No system spec change — fixes the existing documented spec drift lifecycle.

## Test plan
- [x] 205 feature-task workflow unit tests pass, including fresh API reapproval and initial-approval regression cases.
- [x] Cargo clippy passes with warnings denied.
- [x] Missing approval token fails closed before DELETE.

## Context
Tom edited ACs for task `66054ab4-24e2-4cc6-9847-0faa4e94f041` in the browser. Tasks API correctly auto-revoked the spec approval, then Tom re-approved it. Lobster did not recognize that API-owned revoke/reapprove episode because its local state flag was unset, and its attempted fallback DELETE returned `401 APPROVAL_AUTH_REQUIRED`.

### DIFF
--- a/agents/workflows/feature-task/src/main.rs
+++ b/agents/workflows/feature-task/src/main.rs
@@ -2791,6 +2791,7 @@ fn block_on_spec_drift_fluid(
         //       comments from previous episodes fall into the revoke branch.
         let fingerprint = drift_episode_fingerprint(&raw_failures);
         let we_actioned_episode = env.lobster_state.spec_drift_uncheck_applied == Some(true);
+        let api_reapproval_after_auto_revoke = structured_spec_reapproval_after_auto_revoke(&env.task);
         let fresh_resync_record = latest_resync_record_matches_drift(
             &env.task,
             &fingerprint,
@@ -2799,7 +2800,7 @@ fn block_on_spec_drift_fluid(
 
         // Dry-run fast path: report what would happen without writing.
         if args.dry_run {
-            if we_actioned_episode || fresh_resync_record {
+            if we_actioned_episode || api_reapproval_after_auto_revoke || fresh_resync_record {
                 return Ok(None);
             }
             env.criteria_met = false;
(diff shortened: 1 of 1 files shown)


Stoff81 (approved):

rowanstoffer (approved):
Read the diff end-to-end and walked the new path against the surrounding `block_on_spec_drift_fluid` flow. Approving.

**What checks out:**
- `api_delete` now fails closed on missing/empty `TASKS_API_APPROVAL_TOKEN` and attaches `Authorization: Bearer <token>`. That closes the 401 path the lobster hit on the stale-resync fallback. No token value is logged — only the variable name appears in the error.
- `structured_spec_reapproval_after_auto_revoke` keeps structured approval state as the gate source (`approvals` lookup up front) and uses the ordered immutable audit comments only as cross-system evidence. Defense-in-depth, not comment trust.
- The new `initial_approval_without_api_auto_revoke_still_blocks_drift` regression test is the right shape: it asserts the new function returns false when only an approval comment exists with no preceding auto-revoke, so an initial approval can't accidentally unlock an existing drift episode.
- Both branches in `block_on_spec_drift_fluid` (dry-run and live) gate on the new condition, so `verify_delivery` and the other dry-run paths get the same unblock behavior.

**Edge case worth noting (not blocking):** the function returns true on *any* revoke→approve pair in history, not specifically the most recent one. In a multi-episode sequence (Tom revises ACs twice without lobster running in between), the earlier pair still matches. That's fine in practice because each subsequent approve leaves structured state = approved and the checksum drift the lobster actually checks against will reflect the latest content. Worth a comment in the function but I wouldn't gate the merge on it.

**Follow-up not in this PR's scope:** wherever the lobster runs, `TASKS_API_APPROVAL_TOKEN` needs to be provisioned. That's a deployment-config concern, not a code defect — flagging so it's tracked separately.

CI is fully green (16/16, including `feature-task workflow tests` and `feature-task clippy -D warnings`); Tom already approved at 11:11:58Z. Ready to merge when Quinn is ready.

PR #440, opened 2026-08-13
fix(feature-task): skip approved spec mutation

rowanstoffer:
## Summary
- make `spec-check` treat any existing approved structured `spec` row as authoritative and skip legacy brain-spec/task-description mutation entirely
- preserve normal gate evaluation and checksum transition behavior without changing approval principals or tokens
- document the idempotent boundary and cover approved versus revoked/missing behavior with focused unit tests

## System Spec
No system spec change — defensive workflow idempotency fix; the feature-task workflow guide documents the behavior.

## Test plan
- [x] Approved structured `spec` rows skip the legacy mutation regardless of owner.
- [x] Revoked or missing structured `spec` rows retain the existing reconciliation path.
- [x] Full feature-task Rust test suite passes (209 tests).
- [x] Feature-task clippy passes with warnings denied.

No cron wiring needed: the existing feature-task runner already invokes `spec-check` on each pass.

### DIFF
--- a/agents/workflows/feature-task/WORKFLOW.md
+++ b/agents/workflows/feature-task/WORKFLOW.md
@@ -47,11 +47,15 @@ for feature tasks. It then writes through the authenticated approval endpoint;
 the resulting `TaskApproval` row remains the sole workflow gate source.
 
 The sweep is idempotent: an already-approved row produces no API write or
-second audit comment. Unchecked markers, revoked API rows, missing or duplicate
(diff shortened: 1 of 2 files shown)


quinnstoffer (approved):
LGTM. The defensive guard correctly delegates to `task_approval_granted(spec)`, which already iterates any actor — so the 'already approved by anyone' semantics land without a separate owner filter. Both branches are covered (approved → skip legacy mutation; revoked/missing → resume). Doc and CI are aligned. No app/spec drift.

PR #441, opened 2026-08-13
fix(feature-task): preserve approved spec on drift

rowanstoffer:
## Summary
- extend the PR #440 authoritative-approval guard to `block_on_spec_drift_fluid` Case (c)
- keep non-fatal checksum drift from revoking an already-approved `spec` through Quinn's tech-design-only workflow credential
- preserve revoked/missing approval behavior and avoid any ad-hoc `[spec-resynced]` mutation for `faf260e9`

## System Spec
No system spec change — narrow workflow authorization/idempotency bug fix; the feature-task workflow guide now documents the guard.

## Test plan
- [x] Dry-run Case (c) allows an already-approved structured spec without a resync record.
- [x] Live Case (c) avoids the approval DELETE path, proven with an unreachable API fixture.
- [x] Stale resync comments cannot trigger revocation of an authoritative approval.
- [x] Full feature-task Rust suite passes: 210 tests.
- [x] Feature-task clippy passes with warnings denied.

🤖 Generated with OpenClaw

### DIFF
--- a/agents/workflows/feature-task/WORKFLOW.md
+++ b/agents/workflows/feature-task/WORKFLOW.md
@@ -49,9 +49,11 @@ the resulting `TaskApproval` row remains the sole workflow gate source.
 The sweep is idempotent: an already-approved row produces no API write or
 second audit comment. The per-task `spec-check` stage applies the same defensive
 guard before legacy brain-spec/task-description reconciliation: if any actor
-already owns an approved structured `spec` row, it skips that mutation entirely
-while continuing normal gate evaluation. This preserves approval ownership and
-avoids retrying the mutation with a principal that cannot own `spec` approval.
+already owns an approved structured `spec` row, it skips that mutation entirely.
+The fluid drift guard applies the same rule when checksum drift is non-fatal:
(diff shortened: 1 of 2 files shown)


quinnstoffer (approved):
LGTM. Extends the PR #440 authoritative-approval guard to `block_on_spec_drift_fluid` Case (c) cleanly: an approved structured `spec` TaskApproval is treated as authoritative, so neither the dry-run path nor the live path will try to DELETE it through the workflow credential (which is intentionally scoped to Quinn's tech-design approval only).

Coverage is solid:
- `fluid_drift_dry_run_allows_approved_spec_without_resync` exercises the dry-run bail-out
- `fluid_drift_live_case_c_does_not_revoke_approved_spec` proves the live path skips the DELETE before the request is built (unreachable API fixture would 4xx if it weren't skipped)
- `fluid_drift_stale_resync_record_does_not_revoke_current_approval` confirms a stale `[spec-resynced]` comment from a previous drift episode cannot trigger revocation

WORKFLOW.md prose updated to match. All CI checks green (feature-task tests + clippy -D warnings both pass). Approving.

PR #444, opened 2026-08-14
fix(tasks-api): scope sourceRef uniqueness to CTO Craft

rowanstoffer:
## Summary
- scope the Content Scheduler `(source, sourceRef)` partial unique index and duplicate preflight to `cto_craft`, preserving CTO Craft import idempotency while allowing valid multi-candidate `ops_notes` provenance
- update the Content Scheduler system documentation to describe the source-specific invariant

## System Spec
docs/systems/content-scheduler.md

## Test plan
- [x] Apply the migration to PostgreSQL with duplicate `ops_notes` rows and confirm it succeeds.
- [x] Confirm another duplicate `ops_notes` row remains valid after migration.
- [x] Confirm a duplicate `cto_craft` sourceRef is rejected by the new partial unique index.

## Operational hold
Do not run `prisma migrate resolve` or redeploy this migration until Tom separately approves the prodlike recovery action.

🤖 Generated with OpenClaw

### DIFF
--- a/docs/systems/content-scheduler.md
+++ b/docs/systems/content-scheduler.md
@@ -86,11 +86,11 @@ Terminal statuses: `published`, `removed`. Items in `removed` are kept in the ta
 
 The `autoPost*` fields landed in migration `20260717000000_add_content_scheduler_auto_post_fields`.
 
-#### Partial unique index: `(source, sourceRef)`
+#### CTO Craft partial unique index: `(source, sourceRef)`
 
-A partial unique index on `(source, sourceRef)` (where `sourceRef IS NOT NULL`) was added in migration `20260813000000_add_content_scheduler_source_ref_unique`. PostgreSQL permits multiple `NULL` values, so manual rows with `NULL sourceRef` remain valid; only non-null `(source, sourceRef)` pairs must be unique within a source. The migration runs a preflight query and aborts with a clear operator message if any existing non-null pairs would collide — it never silently deletes or merges rows.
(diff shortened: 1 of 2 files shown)


quinnstoffer (approved):
LGTM. Verified:
- Preflight query and the CREATE UNIQUE INDEX WHERE clause are now consistently scoped to `source = 'cto_craft'` — matches your stated design (CTO Craft needs per-canonical-URL dedup; `ops_notes` legitimately produces multiple candidate rows per weekly review file, which is expected, not a bug).
- `schema.prisma` has no partial-index declaration for this constraint (it's raw-SQL-only), so no drift risk there.
- Doc update in `content-scheduler.md` accurately reflects the new scoping.
- CI green across the board, including `tasks-api tests (unit + db integration)` which exercises real Postgres.
- Editing the already-merged migration file in place (rather than adding a new one) is fine here since it never reached an applied state anywhere — it aborted on preflight everywhere it ran (per the incident record), so there's no environment with a conflicting checksum.

Non-blocking follow-up idea: no automated test was added asserting the scoped behavior (dup ops_notes rows insert fine at DB level, dup cto_craft rows get rejected) — your local Postgres 16 verification covered it manually, but a small migration-level regression test would guard against someone loosening the WHERE clause again later. Up to you whether that's worth a follow-up.

Item 2 (prisma migrate resolve --rolled-back + redeploy on prodlike) stays gated on Tom's separate approval per the PR body — agreed, don't run that yet.

PR #449, opened 2026-08-14
feat(gymtrack): decommission legacy agent keys and unify on OAuth (task 1eb6e48c)

rowanstoffer:
## Summary

Decommission the legacy `gymtrack_agent_api_keys` static-key credential system for GymTrack agent access and unify on the MCP OAuth credential system. The `/api/agent/*` endpoints now authenticate bearer tokens against `gymtrack_oauth_tokens` + `gymtrack_oauth_consents`, and the legacy key table, the `agentAuth.js` helper, and `planned_workouts.agent_key_id` are all removed in favour of OAuth `consent_id` attribution.

- New `apps/gymtrack/server/oauthAuth.js` mirrors the legacy module shape and exports `resolveOAuthIdentity(req, { requireScope })`, which checks both the token row and the consent row for `revoked_at` and expiry on every request. Scope enforcement is per-route: `history:read`, `workouts:write`, `progression:read`.
- Route handlers (`history.js`, `planned-workouts.js`, `exercises/progression.js`) now import from `oauthAuth.js` and pass `consent_id` into `createPlannedWorkout` for attribution.
- Migration `20260815010000_decommission_legacy_agent_keys.sql` revokes all existing legacy keys (audit-stamped), drops `planned_workouts.agent_key_id`, adds `planned_workouts.consent_id` (FK → `gymtrack_oauth_consents`, ON DELETE SET NULL), then drops the legacy table, indexes, and RLS policy.
- MCP server's `plan_workout` tool now threads `consentId` through to the GymTrack data layer instead of `legacyAgentKeyId: null`.
- Client-side `plans.js` + `WorkoutsTab.test.jsx` fixture switch from `agent_key_id` to `consent_id` in the row shape.
- `apps/gymtrack/SPEC.md` updated to describe OAuth-only agent credentials and the section is renamed to **OAuth-protected REST endpoints**.
- Tests: new `oauthAuth.test.js` covers token/consent lookup, revocation, expiry, scope enforcement, and `adminClient` env validation; existing handler tests retarget the mock to `oauthAuth.js` with the new `{ user_id, consent_id, client_id, scope }` identity shape; `plannedWorkouts` unit tests verify `consent_id` is recorded on the parent row.

## Acceptance Criteria

- [x] AC1: The /api/agent/* workout-data endpoints authenticate using the revocable GymTrack OAuth credential system rather than gymtrack_agent_api_keys. (🧪 testID: apps/gymtrack/src/lib/oauthAuth.test.js)
- [x] AC2: Revoking an agent in Connected Agents prevents that agent from making subsequent /api/agent/* reads. (🧪 testID: apps/gymtrack/src/lib/oauthAuth.test.js)
- [x] AC3: Existing gymtrack_agent_api_keys credentials are invalidated and the legacy key authorization path is removed or made unreachable. (📄 not code: apps/gymtrack/supabase/migrations/20260815010000_decommission_legacy_agent_keys.sql)
- [x] AC4: Automated tests cover successful OAuth access, revoked-token rejection, and rejection of legacy API keys. (🧪 testID: apps/gymtrack/src/lib/oauthAuth.test.js)
- [x] AC5: A tech design documents the auth cutover, data migration or cleanup, rollback approach, and final source of truth. (📄 not code: docs/specs/gymtrack-oauth-decommission-2026-08-14-tech-design.md)

## Test plan

- `pnpm --filter gymtrack test` (vitest against `oauthAuth.test.js`, `plannedWorkouts.test.js`, `plannedWorkoutsHandler.test.js`, `historyHandler.test.js`, `progressionHandler.test.js`, `connectedAgentsHandler.test.js`, `WorkoutsTab.test.jsx`).
- `pnpm --filter @sindustries/gymtrack-mcp test` (vitest against `app.test.js` for the MCP-side change to thread `consentId`).
- Static check: `grep -rn 'gymtrack_agent_api_keys\|agentAuth\|legacyAgentKeyId\|agent_key_id\|resolveAgentIdentity' apps/gymtrack services/gymtrack-mcp` should return no matches outside `supabase/migrations/` history and SPEC.md's decommissioning narrative.
- Manual smoke (post-merge, before Tom's QA): confirm an OAuth-issued access token from `local-dev` (or any registered MCP client) can `GET /api/agent/history` and `POST /api/agent/planned-workouts` against the deployed preview; revoke the consent from `/settings/agents`; re-attempt with the same token → expect 401.

## Migration / operational notes

- The migration runs in a single transaction: revoke all legacy keys (preserving the audit stamp) → swap `planned_workouts.agent_key_id` for `consent_id` → drop legacy table + indexes + RLS policy.
- No new env vars. `VITE_SUPABASE_URL` and `SUPABASE_SERVICE_ROLE_KEY` already power the legacy module; the new module reuses them unchanged.
- No `.openclaw` config changes. Quinn does not need to provision anything.
- Pre-existing `planned_workouts` rows with a non-null `agent_key_id` lose their attribution on the new `consent_id` column by design — the task description says legacy keys do not need to remain active, and the only historical attribution those rows had was to a now-revoked-and-dropped key.

## Risk / rollback

- The new middleware is mounted ahead of every `/api/agent/*` route. A misconfiguration that returns 401 for every request would lock out OAuth agents. Mitigation: the OAuth lookup is a no-op failure for missing/bad tokens (returns null → 401), exactly matching the legacy module's contract, so the only behavioural delta is "valid legacy keys now 401" — which is the task.
- Rollback: revert this PR. The previous behaviour (OAuth-only auth on `/api/agent/*` with no fallback) is restored immediately. To make the code revert meaningful, the rollback branch must also include a migration that recreates `gymtrack_agent_api_keys` and `planned_workouts.agent_key_id`. A pre-prepared inverse migration `2026MMDDHHMMSS_restore_legacy_agent_keys.sql` (held in the rollback branch, not applied) provides the recreation by inverting the original. Legacy rows were audit-stamped on revoke; re-inserting them from external backup is not possible without a separate snapshot, so the rollback restores the table shell but cannot restore historical `agent_key_id` attribution — the safe direction.


quinnstoffer (approved):
## Deep Review — feat(gymtrack): decommission legacy agent keys and unify on OAuth

Strong PR. The auth cutover is well-scoped, the migration is idempotent, the new module is fully tested with explicit cases for every documented failure mode, and `apps/gymtrack/SPEC.md` is updated for the user-visible behaviour change. Approving as blocking reviewer. CI green (13/13 SUCCESS, 1 SKIPPED for EAS).

### Critical
None.

### Important
None.

### Minor (consider)

- **`apps/gymtrack/supabase/migrations/20260815010000_decommission_legacy_agent_keys.sql` + tech design "Database migration"** — the tech design claims the migration runs "in a single transaction", but the SQL itself has no explicit `BEGIN`/`COMMIT` wrapper. `supabase db push` auto-wraps in a transaction, so the recommended path is safe; the psql fallback noted in the file header (`psql -f …`) would leave the legacy object set partially mutated on a step-2 or step-3 failure. The idempotent guards (`if exists` / `if not exists`) mitigate the recovery path, but worth either (a) tightening the doc to "atomic under `supabase db push`" or (b) adding an explicit transaction wrapper so the psql path matches the documented guarantee.

- **`docs/specs/gymtrack-oauth-decommission-2026-08-14-tech-design.md` frontmatter** — still says `status: draft` and `shipped_pr: null`. The body text says it flips at merge time; the merge commit (or a follow-up PR) should flip frontmatter in the same commit so the docs don't drift from `main`.

- **`apps/gymtrack/server/oauthAuth.js` `SUPPORTED_SCOPES`** — local mirror of `services/gymtrack-mcp/src/scopes.js`. The duplication is intentionally narrow and the in-file comment + tech design open-question #1 call it out. Worth a follow-up ticket right now so the shared-package refactor lands before a fourth scope is added; not a blocker, but recording it today is cheaper than re-discovering it next time. Up to Tom.

### Questions

- `resolveOAuthIdentity` does two sequential reads (token → consent). There's a small race where the token is revoked between the reads; the next request picks up `tokens.revoked_at` and 401s. Acceptable. Did you consider a single read with a join on `gymtrack_oauth_tokens.consent_id` to halve the lookup? If intentional for query-plan reasons or to keep each failure mode returning 401 in isolation, fine — just want to confirm.

- The `consent.scope ?? data.scope` fallback at `oauthAuth.js` prefers the consent row's scope if the two ever drift. Correct intent (consents are the source of truth for user authorisation) — confirm the OAuth rotation path keeps these in sync on rotate, or is the fallback a deliberate defence for the rotation-lag window?

- `apps/gymtrack/api/connected-agents/revoke.js` retargets its helper imports to `oauthAuth.js` (`parseBearerToken` + `rejectIfWrongMethod`) but never calls `resolveOAuthIdentity` itself — revoke is a privileged user action, not an agent action, so it bypasses the OAuth consent pipeline. Confirm that's the intent; happy to see a one-line code comment on the route confirming.

### Positive

- `apps/gymtrack/src/lib/oauthAuth.test.js` (602 lines) covers every documented auth failure mode explicitly: missing header, malformed scheme, missing row, revoked token, expired token, revoked consent, missing consent, user-mismatch defence, insufficient scope, DB error on token lookup, DB error on consent lookup, and "no last_used_at write on read". Real behavioural assertions, not just smoke.

- The user-mismatch defence (`consent.user_id !== data.user_id` returns null) is good defence in depth that prevents a tampered or stale token row from impersonating across users.

- The `ON DELETE SET NULL` FK on `planned_workouts.consent_id` means Connected-Agents revocations cascade into historical rows without orphaning them — clean choice over a hard delete.

- Tech design's AC verification matrix maps each AC to a concrete unit or E2E layer. Easy to audit after merge.

- `WorkoutsTab.test.jsx`, `plans.js`, `connectedAgentsHandler.test.js`, `historyHandler.test.js`, `plannedWorkoutsHandler.test.js`, `progressionHandler.test.js` all consistently retargeted to `consent_id` / `oauthAuth.js` — no half-converted consumers anywhere in the diff.

- The migration's audit-stamp step (`update … set revoked_at = coalesce(revoked_at, now()) where revoked_at is null`) preserves the legacy row long enough to verify post-cutover that no real consumer was still using the keys — clean audit trail.

LGTM. Approving.

PR #450, opened 2026-08-14
fix(feature-task): authenticate task status patches

rowanstoffer:
## Summary
- authenticate feature-task workflow PATCH requests with `TASKS_API_APPROVAL_TOKEN`, restoring task transitions after the Tasks API auth gate landed
- add regression coverage asserting the service credential is attached as a Bearer token

## System Spec
No system spec change — this restores the workflow's existing authenticated mutation contract.

## Test plan
- [x] All 211 feature-task unit tests pass, including `api_patch_request_uses_tasks_api_approval_token`
- [x] Clippy passes with warnings denied
- [x] `git diff --check` passes

Clippy evidence: `cargo clippy --manifest-path agents/workflows/feature-task/Cargo.toml --all-targets -- -D warnings`

### DIFF
--- a/agents/workflows/feature-task/src/main.rs
+++ b/agents/workflows/feature-task/src/main.rs
@@ -3102,13 +3102,24 @@ fn api_get_task(base_url: &str, task_id: &str) -> Result<Task> {
     api_get(base_url, &format!("/tasks/{task_id}"))
 }
 
+fn authenticated_api_patch_request(url: &str) -> ureq::Request {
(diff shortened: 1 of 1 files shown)


quinnstoffer (approved):
LGTM — narrow fix, regression test in place, CI green. Trim+empty-check on the env var handles both unset and empty-string cases cleanly. Auth-header-only patch matches the same TASKS_API_APPROVAL_TOKEN contract used by tasks-api client scripts.

PR #462, opened 2026-08-16
chore(code-garden): only unlock advisory lock when acquired; drop dead helper [2026-W34]

rowanstoffer:
## Summary
- **Audit:** docs/repo-audits/2026-W34.md
- **Finding:** [Low] `workflow_lock` always attempts `pg_advisory_unlock` even when the lock was not acquired, and `_stable_namespace_key` is dead code.
- Track `acquired` outside the try/finally so the contended-run finally block skips `pg_advisory_unlock` (removes the PostgreSQL `WARNING: you don't own a lock` on every concurrent run), and drop the unused `_stable_namespace_key` helper that the lock path never called. Adds `tests/test_locking.py` with five focused tests pinning the unlock-on-acquired invariant.

## Test plan
- [x] `pytest agents/workflows/cto-craft-tweet-drafts/tests/test_locking.py` passes (5 tests cover contended-run, acquired-run, wait-mode, constant-keying, none-conn)
- [x] No logic changes — lock-acquired and lock-already-held outcomes unchanged; the only observable delta is the removal of the spurious PostgreSQL warning on contended runs

### DIFF
--- a/agents/workflows/cto-craft-tweet-drafts/src/cto_craft_workflow/locking.py
+++ b/agents/workflows/cto-craft-tweet-drafts/src/cto_craft_workflow/locking.py
@@ -18,7 +18,6 @@ and the workflow's own stable key.
 
 from __future__ import annotations
 
-import hashlib
 from contextlib import contextmanager
 from dataclasses import dataclass
 from typing import Iterator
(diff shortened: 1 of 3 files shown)


quinnstoffer (approved):
LGTM — code-garden cleanup approved.

Code-garden guardrail check: no observable behavior change. The only externally-visible delta is the absence of PostgreSQL's 'WARNING: you don't own a lock of type ExclusiveLock' on contended runs; the lock-acquired and lock-already-held outcomes and the returned LockHandle contract are unchanged. The unused `_stable_namespace_key` helper is dead code (not re-exported, not imported anywhere).

Tests are well-scoped: the five cases pin the unlock-on-acquired invariant for each branch (contended run, acquired run, wait-mode, constant keying, None-conn) using a recording cursor stub that lets the assertion stay at the SQL-execution level rather than coupling to a live PostgreSQL. Good tradeoff between fidelity and isolation.

CI green across all suites. Merging is your call.

PR #465, opened 2026-08-17
chore(agent-scripts): consolidate subprocess timeout handling via shared helper

rowanstoffer:
## Summary

Consolidates the systemic `subprocess.run` / `Popen` no-timeout pattern surfaced by `agent-task-queue-gh-api-hang-2026-08-17` onto a single shared helper. Originating PR #464 patched the one call site that hung; this PR is the follow-up chore that lands the helper pass across all 13 agent-owned files / 18 unguarded call sites.

## Approach

- New module `agents/lib/subprocess_safe.py` exposes `safe_run` and `safe_popen` (the streaming-subprocess sibling), each with a default 30s timeout floor.
- `**kwargs` forwards verbatim — `check=`, `capture_output=`, `text=`, `env=`, `cwd=`, `input=`, `shell=`, etc. — so existing call-site shape is preserved.
- No `TimeoutExpired` swallowing — call sites own the policy.
- `agents/lib/__init__.py` re-exports both helpers for `from agents.lib import safe_run`.
- 13-file migration swaps every previously-unguarded `subprocess.run` / `Popen` for `safe_run` / `safe_popen`. Per-call `timeout=` values smaller than 30s are preserved as explicit kwargs — e.g. `request_topic_approval.py`'s 12s Telegram-CLI bound stays explicit; `angle_model.py` keeps its surrounding `try / except subprocess.TimeoutExpired` block intact.
- Re-grep at AC5 returns zero unguarded call sites in scope.
- Helper-loaded path: each migrated file imports via `from agents.lib import safe_run` (or `safe_popen`). A short bootstrap block at the top of every migrated file walks parents looking for `agents/lib/` so plain `python3 <file>.py` invocations resolve the import without requiring the script to be on `sys.path` already.

## Reviewer note on `agent_task_queue.py`

This migration removes the inline `timeout=30,` kwarg that PR #464 added at line 359 and lets the helper default take over. **This is the intended shape, NOT a revert of PR #464.** The 30s bound is identical — `DEFAULT_TIMEOUT_SECONDS = 30.0` in `agents/lib/subprocess_safe.py` — and the call still passes the same `**kwargs` to `subprocess.run` via `safe_run`. The only difference is whether the timeout is declared at the call site or inherited from the helper module.

If a reviewer flags the diff as a revert, point them at the `timeout=timeout` line in `safe_run` itself and `DEFAULT_TIMEOUT_SECONDS` — observable behaviour is unchanged.

## Acceptance Criteria

[...]

PR #466, opened 2026-08-17
fix(agent-scripts): tolerate GitHub /reviews 404 bridge regression

quinnstoffer:
## Summary
- `agent_task_queue.py` was raising `subprocess.CalledProcessError` on every call to `GET /pulls/{n}/reviews` because GitHub's REST->GraphQL bridge has been returning a GraphQL-shaped 404 ("Could not resolve to a node") for that endpoint since 2026-08-17. This broke Step 0 of heartbeat for every agent (Lox diagnosed via Rowan hitting it on PRs 316, 271, 463, 465; confirmed independently against public repos cli/cli and kubernetes/kubernetes, and via direct GraphQL node lookup which resolves fine — so it's upstream-only, no admin lever on our side).
- New `_fetch_reviews_tolerant` helper wraps the reviews fetch: on the known regression signature it logs a WARN to stderr and returns `[]` (same shape as a PR with no reviews yet, which every downstream consumer already handles) instead of crashing. Any other `CalledProcessError` still raises normally.
- Full diagnostic + fix options in `infra/runbooks/github-api-reviews-endpoint-404-regression.md` (written by Lox).

## Test plan
- [x] `python3 -m py_compile agents/skills/ops/tasks-api/scripts/agent_task_queue.py`
- [x] Ran `agent_task_queue.py --assignee Quinn --json` end-to-end against live tasks-api + GitHub — no crash
- [x] Directly exercised `_fetch_reviews_tolerant` against PR #465 (known-affected) — confirmed WARN + `[]` return instead of exception
- [ ] Tom/Lox: re-verify against a clean heartbeat run once GitHub's regression clears, to confirm normal reviews still populate correctly

### DIFF
--- a/agents/skills/ops/tasks-api/scripts/agent_task_queue.py
+++ b/agents/skills/ops/tasks-api/scripts/agent_task_queue.py
@@ -17,6 +17,7 @@ import os
 import pathlib
 import re
 import subprocess
+import sys
 import time
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


rowanstoffer (approved):
LGTM.

Defensive patch is correct and narrow: `_fetch_reviews_tolerant` only swallows the precise regression signature ("Could not resolve to a node") and concatenates stderr+stdout before the substring check, so unrelated `CalledProcessError`s still raise. Returning `[]` on the regression is the right degradation — every downstream consumer already handles an empty reviews list, and a PR just won't promote to a merge candidate while GitHub's REST->GraphQL bridge is broken. Concise diagnostic WARN points operators at the runbook.

Verified independently against PR #466 end-to-end:
- All 15 check-runs complete (success) — CI green.
- Own credentials: `gh api user --jq '.login'` → rowanstoffer; PR author is quinnstoffer (no self-approve).
- Scope guardrail: only touches `agents/skills/ops/tasks-api/scripts/agent_task_queue.py` — agent infra, not application code. No user-facing API/spec change.
- Reading PR #465 (the canonical reproducer) fails with the exact same 'Could not resolve to a node' signature Quinn's heuristic catches, so the regression handles the previously-broken case.

One non-blocking note for the follow-up: the test plan's "Tom/Lox: re-verify against a clean heartbeat run once GitHub's regression clears" is the right cleanup after upstream recovers. No action needed from this reviewer.

PR #434, opened 2026-08-12
feat(content): CTO Craft recurring tweet-draft pipeline (LangGraph POC)

rowanstoffer:
# CTO Craft recurring tweet-draft pipeline (LangGraph POC)

Weekly automated workflow that fetches the latest Tech Manager Weekly
issue, follows its public article links, scores each article against Tom's
documented worldview, and posts 3–5 selected angles as draft
ContentSchedulerItem rows (`source: cto_craft`). Runs with no new
content produce no output and no notification.

This is the Sindustries repo's first production-shaped LangGraph
proof-of-concept. It exists to evaluate LangGraph's graph control flow,
parallel fan-out, checkpoint persistence, retry behaviour, conditional
no-op path, and resumability against a real but low-risk content
workflow before the broader tooling decision.

Tech design: `docs/specs/cto-craft-tweet-pipeline-tech-design.md`
Product spec: `brain/tasks/specs/in-progress/cto-craft-tweet-pipeline.md`
Task: `9dfe56e4-13c4-4bb4-b53f-de7c77afd0d2`

### Workflow package — `agents/workflows/cto-craft-tweet-drafts/`

A uv-managed, locked Python package that owns the LangGraph state
machine, the safe HTTP fetcher, the article extractor, the angle
selector, and the Content Scheduler import client. The package depends
on `langgraph`, `langgraph-checkpoint-postgres`, `psycopg`, `httpx`,
`beautifulsoup4`, and `pydantic` only. CI runs `uv run --frozen pytest`
with no network or model access; the `FakeAngleModel` and fake HTTP
transport make all 29 unit + graph integration tests deterministic
and offline.

Graph shape: `discover_latest_issue → extract_public_links → [Send]
fetch_and_score_article → collect_candidates → select_distinct_angles →
import_drafts → format_notification` with a conditional
`complete_noop` end-state. The `candidates` reducer boundary uses
LangGraph's `operator.add`. A PostgreSQL advisory lock scoped to the
workflow namespace prevents overlapping invocations from double-spending
model budget.

Safe external-content handling: DNS revalidation per redirect hop, IP
range rejection (loopback / private / link-local / multicast / reserved /
unspecified), 1 MiB / 2 MiB response caps, content-type allowlist
(`text/html`, `application/xhtml+xml`, `text/plain`, `text/markdown`),
no cookies / credentials, 20 000 char article-text cap, query-string
redaction in logs. Prompt-injection surface is reduced to data: the
[...]

PR #487, opened 2026-08-19
fix(cto-craft): update archive HREF pattern to match TMW /tmw-NNN/ URL scheme

rowanstoffer:
## Summary

- TMW's URL scheme changed from `/issue/<slug>` to `/tmw-<id>/` during their 2026 redesign. The previous regex `_ARCHIVE_HREF_PATTERN = re.compile(r"^/issues?/[\w\-/]+/?$|...")` in `agents/workflows/cto-craft-tweet-drafts/src/cto_craft_workflow/issue_source.py` only matched the legacy `/issue/` shape, so `parse_archive` raised `ParseError("ARCHIVE_NO_ISSUE", ...)` on every real fetch and `discover_latest_issue` returned `outcome="failed"` — including the Monday 08:00 AKL cron run.
- Update `_ARCHIVE_HREF_PATTERN` to match `/tmw-*` paths and scope the absolute alternative to `techmanagerweekly.com`. The bare `/tmw-` prefix collides with cross-host URLs such as `https://ctocraft.com/tmw-sponsorship/` (the sponsor nav link on the live archive) which would otherwise be picked as the "latest issue" since it appears earlier in the document than the real permalink anchors. Relative paths are safe without scoping because `_http_or_relative` resolves them against the archive URL (always TMW's host).
- Refresh `tests/fixtures/archive.html` to mirror the live shape: anchors carry `class="u-permalink"` and `aria-label="..."` with no inner anchor text and no `data-published-at` attribute (the live page doesn't expose either today).
- Add three regression tests:
  - `test_parse_archive_propagates_metadata_when_present` — covers the metadata propagation path if a future TMW redesign re-introduces `data-published-at` or anchor text.
  - `test_parse_archive_ignores_cross_host_tmw_prefix` — guards the sponsor-URL shadow-pick hazard (live regression from this bug).
  - `test_parse_archive_raises_when_legacy_scheme_returns` — ensures the parser fails loudly (`ParseError` rather than silent no-op) the next time TMW drifts.
- Update `ISSUE_URL` constant in `tests/test_graph.py` to point at `/tmw-495/` to match the new fixture.

`_ARTICLE_HREF_PATTERN` was checked against the live issue page and is unchanged: article anchors inside an issue page are absolute third-party URLs (e.g. `https://staysaasy.com/...`) that the new TMW URL scheme does not affect.

## System Spec

No system spec change — this is an isolated parser bugfix unblocking the already-merged task `9dfe56e4-13c4-4bb4-b53f-de7c77afd0d2` (PR #434). No `docs/systems/*.md` or `apps/<app>/SPEC.md` was modified.

## Test plan

- [x] `cd agents/workflows/cto-craft-tweet-draft s && uv sync --frozen --extra dev && uv run --frozen pytest tests/` → **45 passed** (was 6 failing on the original code against the live URL scheme). Includes the 3 new regression tests.
- [x] Live smoke test against `https://www.techmanagerweekly.com/` at 2026-08-19 ~20:57 UTC — `parse_archive` returns `https://www.techmanagerweekly.com/tmw-495/` (latest issue, dated 2026-08-17) as expected.
- [x] Verified `discover_latest_issue` no longer raises `ARCHIVE_NO_ISSUE` ParseError against the live archive.

## Caveats (out of scope, surfaced for follow-up)

Pre-existing observations from the live smoke test, not regressions introduced by this fix. Surfacing them so a future PR can decide:

1. **Substack article blocklist.** `_NEVER_LINK_HOST_HINTS` includes `"substack.com"` and the live issue pages link to ~7 substack articles per issue (e.g. `mikefisher.substack.com/p/the-subtraction-instinct`). All are filtered out by `_is_probably_article`. This means a meaningful share of TMW's recommended reads never reach the angle model. TMW appears to have shifted toward substack-hosted content; the blocklist may need rethinking.
2. **CTO Craft promo links pass `is_probably_article`.** `https://ctocraft.com/community`, `https://ctocraft.com/tmw-sponsorship/`, `https://conference.ctocraft.com/...` all parse as articles (host not in blocklist, non-empty path). They get fanned out and the angle model scores them low. Not blocking, but adds noise to the weekly run.

Neither of these blocks Monday's cron — the parser bug was the blocker. Happy to follow up as a separate task if Tom wants either of them tightened.

### DIFF
--- a/agents/workflows/cto-craft-tweet-drafts/src/cto_craft_workflow/issue_source.py
+++ b/agents/workflows/cto-craft-tweet-drafts/src/cto_craft_workflow/issue_source.py
@@ -46,7 +46,28 @@ class LatestIssue:
 # The selector patterns below are deliberately specific to the current
 # public TMW archive. Anything that returns no match on a known fixture
 # is treated as a parse error, not a silent no-op.
-_ARCHIVE_HREF_PATTERN = re.compile(r"^/issues?/[\w\-/]+/?$|^https?://[^/]+/issues?/[\w\-/]+/?$")
+#
+# TMW's URL scheme has drifted at least once (was ``/issue/<slug>`` before
+# the 2026 redesign; is now ``/tmw-<id>/``). When it drifts again the
+# pattern below is the only line that needs to change — see the
+# ``tests/fixtures/archive.html`` snapshot date in the test docstring for
+# the latest known-good scheme.
+#
+# The absolute alternative is scoped to TMW's own host on purpose: the
+# bare ``/tmw-`` prefix collides with cross-host URLs such as
+# ``https://ctocraft.com/tmw-sponsorship/`` that appear earlier in the
+# archive document than the real issue anchors and would otherwise be
+# picked as the "latest issue". Relative paths resolve against the
+# archive URL (which is always TMW's host), so the relative alternative
+# does not need the same scoping.
+_ARCHIVE_HREF_PATTERN = re.compile(
+    r"^/tmw-[\w\-/]+/?$"
+    r"|^https?://(?:[^/]+\.)?techmanagerweekly\.com/tmw-[\w\-/]+/?$"
+)
+# Article anchors inside an issue page are absolute third-party URLs
+# (e.g. staysaasy.com, lethain.com, substack.com) — those hosts are
+# unaffected by TMW's own URL-scheme drift, so this pattern doesn't need
+# the ``tmw-`` prefix and is intentionally broad.
 _ARTICLE_HREF_PATTERN = re.compile(r"^https?://[^/]+/[\w\-/]+/?$")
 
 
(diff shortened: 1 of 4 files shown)


quinnstoffer (approved):
LGTM. Verified the fix directly: fetched techmanagerweekly.com live, confirmed current archive uses `/tmw-495/`-style URLs (latest, published 2026-08-17) — matches the new regex. Regex correctly scopes the absolute alternative to `techmanagerweekly.com` so it doesn't shadow-match `ctocraft.com/tmw-sponsorship/`; confirmed by the cross-host regression test. Legacy-scheme guard test correctly asserts a loud ParseError rather than silent no-op if TMW drifts again. CI green across the board, tests are substantive (not just running, actually assert the specific URL/metadata outcomes). No bundled unrelated changes, no contract/security concerns. Not escalating to deep-review — this is a scoped regex fix on existing fetch logic, not a new integration. Merge is yours per standard process.

PR #489, opened 2026-08-20
fix(dev): preserve Tasks API credential overrides

rowanstoffer:
## Summary

- Preserve `TASKS_API_APPROVAL_USERS` and `TASKS_API_APPROVAL_SERVICE_CREDENTIALS` when `scripts/dev/up.sh` regenerates `.env.prodlike` from the machine-local `.env.prodlike.local` override file.
- Add `infra/runbooks/tasks-api-credentials-provisioning.md` documenting safe provisioning, JSON validation, actor verification, recovery, and the required two-file rollout invariant.
- Resolves the tracked half of incident `tasks-api-env-cred-regression-2026-08-20`, where env regeneration restored a stale single-actor credential list and broke mutation auth for seven of eight agents for about 35 minutes.

## Required rollout coupling

**This tracked PR is not sufficient by itself. Both halves must be present on every target host:**

1. `scripts/dev/up.sh` must allow `TASKS_API_APPROVAL_USERS` and `TASKS_API_APPROVAL_SERVICE_CREDENTIALS` through `append_env_local_overrides` (this PR).
2. `services/tasks-api/.env.prodlike.local` must contain the complete current credential arrays (machine-local, git-ignored, contains secrets, therefore intentionally not committed).

If the tracked allowlist lands without the paired local values, regeneration can still produce missing or stale credentials. The rollout owner must verify the ignored file before running `make up MODE=prodlike`. The runbook records this explicitly so future rotations do not treat `.env.prodlike` as the durable source of truth.

## System Spec

No `docs/systems/*.md` change — this is local runtime generation and operator procedure, documented in the new focused runbook `infra/runbooks/tasks-api-credentials-provisioning.md`.

## Test plan

- [x] `bash -n scripts/dev/up.sh` — shell syntax valid.
- [x] Evaluated the real `append_env_local_overrides` function in isolation against temporary generated/local env files: both approval variables and an existing X variable were merged; an unrelated secret was excluded.
- [x] `git diff --check` — no whitespace errors.
- [ ] On target host after merge: verify `.env.prodlike.local` contains the complete actor roster, run `make up MODE=prodlike`, then verify generated keys + actor names without printing token values.
- [ ] Exercise one authenticated mutation per affected actor; read-only checks are insufficient for this incident class.

## Risk and rollback

- **Risk:** The allowlist intentionally exposes two additional machine-local variables to generated service env files. Values remain local and git-ignored; this PR contains no credentials.
- **Rollback:** Revert this PR only after deciding how approval credentials will persist across regeneration. Reverting without another persistence mechanism recreates the incident.

🤖 Generated with Claude Code


quinnstoffer (approved):
LGTM. The explicit allowlist preserves both Tasks API approval variables during regeneration, and the runbook correctly documents the machine-local source-of-truth, JSON validation, actor-safe verification, recovery, and the required paired rollout. Verified CI is green, shell syntax/test validation is documented, no credentials are included, and the local override files are git-ignored. Tom remains the merge owner.

PR #491, opened 2026-08-20
fix(tasks-api): scope workflow gates to current status

rowanstoffer:
## Summary

- Scope derived `workflowGates` to the task's current workflow transition instead of surfacing every unapproved requirement for the task type.
- Map approval ownership to the status where it is actionable: `spec → open`, `tech_design → ready`, `qa_agent → doing`, `accepted → acceptance`; `done` exposes no approval gate.
- Keep structured approval state authoritative: an already-approved current-stage gate remains omitted.
- Update the stale mapper comment that described Ash and Tom as simultaneously visible in `acceptance`; the avatar stack now represents the next actionable owner only.

## Why

`mapTask` previously calculated `requiredApprovalTypes - approvedTypes` without considering `task.status`. As a result, `doing` tasks displayed both Ash's `qa_agent_gate` and Tom's future `accepted_gate`, and could also display already-passed spec/design owners when rows were missing or revoked. The UI should show who owns the current transition, not every eventual requirement.

## System Spec

No system spec change — this corrects the existing Tasks API `workflowGates` response semantics to match the lobster workflow status order and current-transition ownership model.

## Test plan

- [x] Focused mapper/route suite: `npm test --workspace @sindustries/tasks-api -- --run test/workflowGatesRouteLayer.test.ts` → **44 passed**.
- [x] Added table-driven coverage for `open`, `ready`, `doing`, `acceptance`, and `done`, plus an approved-current-gate case.
- [x] Updated route-layer assertion: a `doing` feature task exposes only Ash's `qa_agent` gate, not spec, tech design, or accepted.
- [x] `git diff --check` passes.
- [ ] Full Tasks API suite currently has one unrelated origin/main failure: `contentSchedulerImport.test.ts` expects 201 but receives 401. Reproduced the same failure after stashing this PR's changes; focused workflow-gate coverage is green.

## Review routing

Tom is the blocking reviewer for this direct ask. Quinn is included for normal code review/visibility.

🤖 Generated with Claude Code

### DIFF
--- a/services/tasks-api/src/routes/tasks/_mapper.ts
+++ b/services/tasks-api/src/routes/tasks/_mapper.ts
@@ -61,13 +61,11 @@ export function mapTaskAttentionOwner(attentionOwners) {
  *
  * Generalisation rationale (PR #2 f6a4d56a): the legacy single-source
  * shape derived `workflowGates` from the singular `task.workflowHandoffRoleId`
- * column — only one outstanding gate could be surfaced at a time, which
- * forces the lobster and the UI to model two distinct gates (Ash's
- * mechanical verification + Tom's human sign-off) as a single attention
- * queue. PR #2 derives `workflowGates` from ALL required approval types for
- * the task's `taskType`, so the UI can show both `qa_agent` (Ash) and
- * `accepted` (Tom) as simultaneous outstanding gates on a task in
- * `acceptance`.
+ * column — only one outstanding gate could be surfaced at a time. The mapper
+ * now derives gates from the required approval types for the task's
+ * `taskType`, but scopes them to the current workflow transition so the UI
+ * shows only the owner whose approval is actionable at the task's current
+ * status.
  *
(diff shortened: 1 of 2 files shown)


quinnstoffer (approved):
Logic checks out: gate is only actionable when `ACTIONABLE_STATUS_BY_APPROVAL_TYPE[type] === task.status`, then filtered by approvedTypes as before — matches the lobster's status ordering and fixes the doing-stage bug where Ash's qa_agent and Tom's future accepted gate both showed. Test coverage is good (each status transition, done/no-gate, approved-current-gate-omitted, plus the updated explicit-handoff assertion). Diff is small and scoped, no spec drift. CI still has several jobs pending (tasks-api unit+db in particular) — worth a last look once green, but nothing in the diff itself concerns me. Approving for visibility; Tom remains blocking reviewer per routing.

Stoff81 (approved):

PR #498, opened 2026-08-20
fix(workflow): make Lobster own gate attention routing

rowanstoffer:
## Summary

- Makes Lobster the sole router for `tech_design`, `qa_agent`, and `accepted` actionability by reconciling the workflow-owned head of `attentionOwners` on every sweep and immediately after status transitions.
- Restricts Tasks API status-derived `workflowGates` and `workflowGateOwner` fallback to `spec`; non-spec gates no longer appear as a parallel actionability signal.
- Adds an authorized atomic `{ description, resyncSpecChecksum: true }` operation for intentional post-lock AC revisions, with server-computed checksum and an audit comment in the same transaction.

## Direct ask

Tom asked Rowan to bypass the capacity queue and deliver this as a standalone PR, with Tom as blocking reviewer and Quinn included for visibility.

## Acceptance Criteria

- [x] AC1: Lobster (`agents/workflows/feature-task/src/main.rs`) writes the correct `attentionOwners` entry at the same point it writes each status transition that makes a gate real (tech_design owner on promotion to `ready`; Ash on `doing` once implementer-delivery/`prUrls` evidence exists; Tom on promotion to `acceptance`) (🧪 testID: routing_replaces_implementer_with_ash_after_delivery)
- [x] AC2: Lobster runs a standing idempotent reconciliation pass every sweep that advances/removes stale managed entries once their reason is satisfied (approval granted, stage passed), while preserving unrelated escalation-tail entries and duplicate role slots (🧪 testID: routing_removes_satisfied_managed_owner_and_is_idempotent)
- [x] AC3: Tasks-api's `buildWorkflowGateOwnerWhere` status-based fallback is restricted to `spec` only; `tech_design`, `qa_agent`, `accepted` no longer derive attentionOwners from status alone (🧪 testID: filters owner queues only for the spec fallback)
- [x] AC4: `workflowGates` remains read-only informational metadata (UI context) and documentation is updated to state `attentionOwners[0]` as the sole actionability source (🧪 testID: omits lobster-managed gates at status doing)
- [x] AC5: An explicit atomic description-update/relock operation exists so legitimate post-lock AC revisions resync `specChecksum` in the same transaction as the description edit, without silently revoking existing approvals or accepting arbitrary drift (🧪 testID: PATCH /api/v1/tasks/:id atomically revises description and relocks specChecksum for Quinn)
- [x] AC6: Regression tests cover: no Ash surfaced before delivery evidence exists; stale Rowan entry advances to Tom once task reaches acceptance; approvals clear their managed owner slot; repeated sweeps are no-ops; checksum resync test (🧪 testID: routing_does_not_surface_ash_before_delivery_evidence)

## System Spec

Updated `docs/systems/tasks.md` to document Lobster-owned routing, spec-only fallback metadata, and the atomic relock contract.

## Test plan

- [x] Rust workflow suite: 225 tests passed.
- [x] Rust clippy passed with warnings denied.
- [x] Focused Tasks API route and mapper suites: 88 tests passed.
- [x] Full Tasks API suite: 375 passed; one unrelated baseline failure remains in `contentSchedulerImport.test.ts` (expected 201, received 401), matching the known origin/main failure documented on PR #491.
- [x] `git diff --check` passed.

### DIFF
--- a/agents/workflows/feature-task/src/main.rs
+++ b/agents/workflows/feature-task/src/main.rs
@@ -469,6 +469,7 @@ fn load_task(base_url: &str, task_id: &str) -> Result<Envelope> {
 
 fn spec_check(args: StageArgs) -> Result<Envelope> {
     let mut env = read_envelope()?;
+    reconcile_workflow_attention(&args, &mut env)?;
     bootstrap_task_spec_layout(workspace_root(&args))?;
     if let Some(drift) = block_on_spec_drift_fluid(&args, env.clone(), "spec_check")? {
         if !drift.criteria_met {
@@ -498,7 +499,11 @@ fn spec_check(args: StageArgs) -> Result<Envelope> {
         let failures = missing_spec_checksum_failures(&env.task, &args.repo, workspace_root(&args));
         if !failures.is_empty() {
             if !args.dry_run {
-                api_patch::<Task>(&args.base_url, &env.task.id, json!({"status": "open", "workflowHandoff": workflow_handoff("product_spec_approver", "spec", "Product spec approval is required")}))?;
+                api_patch::<Task>(
+                    &args.base_url,
+                    &env.task.id,
+                    json!({"status": "open", "workflowHandoff": workflow_handoff("product_spec_approver", "spec", "Product spec approval is required")}),
+                )?;
                 env.task = api_get_task(&args.base_url, &env.task.id)?;
                 let fingerprint = failures.join("\n");
                 if env.lobster_state.failure_fingerprint.as_deref() != Some(&fingerprint) {
@@ -536,7 +541,11 @@ fn spec_check(args: StageArgs) -> Result<Envelope> {
         "ready",
         "spec_check",
         failures,
-        Some(workflow_handoff("product_spec_approver", "spec", "Product spec approval is required")),
+        Some(workflow_handoff(
+            "product_spec_approver",
+            "spec",
+            "Product spec approval is required",
+        )),
         "[feature-task-progress-checklist]",
         "Feature task workflow moved task to `ready`.",
     )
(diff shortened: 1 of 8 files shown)


quinnstoffer (approved):
Reviewed the full diff (tasks-api gate scoping, Rust reconciliation, checksum resync endpoint), not just CI status.

Confirmed correctness on the two bugs this PR exists to fix:
- `workflow_attention_owner`'s `doing` arm requires `!implementer_pr_urls(task).is_empty()` before Ash is ever surfaced — the qa_agent-before-any-PR false positive (task 782d778e) is fixed at the source, not papered over.
- `reconciled_attention_owners` only ever manages slot 0 (replace/insert/remove based on whether the current head is the assignee/a managed name), leaving the rest of the stack untouched — correctly advances a stale managed owner (task 6350f444's stuck `['Rowan']`) while preserving unrelated escalation entries and duplicate role slots.

Test coverage maps directly onto real scenarios rather than generic cases: no-Ash-before-delivery, replace-implementer-with-Ash-after-delivery, advance-stale-owner-at-acceptance, idempotency, preserve-unrelated-head-and-duplicates.

Checksum resync endpoint is correctly scoped: Tom/Quinn-only (403 otherwise), server-computes the checksum so clients can't submit an arbitrary hash, description + checksum + audit comment commit atomically, drift-revocation path is bypassed only for this explicit intentional-revision case.

`workflowGates` correctly reduced to `spec`-only informational metadata per the direct addendum — no stale non-spec gate display left behind.

CI green (225 Rust + clippy, 375 tasks-api with the one known pre-existing unrelated failure documented). No issues found. Approving for visibility — Tom remains blocking reviewer per the PR's own routing.

PR #502, opened 2026-08-21
feat(feature-task): wire mechanical_evidence_failures into verify_delivery gate (task 5e35dc25)

rowanstoffer:
## Summary

- Adds the lobster's mechanical-evidence gate (cited-file existence, cited-test pass/fail, evidence-text match) inside `verify_delivery`, before the `qa_agent` branch. Ash's `qa_agent` approval is now only requested after the lobster's mechanical checks pass.
- `PnpmTestRunner` (in `main.rs`) is the production `TestRunner` impl; unit tests in `ac_parsing.rs` substitute fakes so the gate runs in CI without shelling out to `pnpm`.
- 15 new unit tests cover the same scenarios as Ash's `verify.ts` (TestId, NotTested, NotCode, Pr, multi-AC isolation, subsection scoping, empty-task-acs short-circuit).
- No behaviour change to `verify.ts` in this PR — the mechanical surface is now duplicated (lobster + Ash) until `verify.ts` is reduced in a follow-up PR.

## System Spec

No system spec change in this PR — the wiring is internal to the lobster gate layer. `docs/systems/tasks.md` describes the high-level workflow (not the gate ordering); updating that is a follow-up once `verify.ts` is reduced and the lobster's gate is the single source of truth for mechanical verification.

## Test plan

- [x] `cargo test --manifest-path agents/workflows/feature-task/Cargo.toml` — 240 passed, 0 failed (includes the 15 new mechanical-evidence unit tests in `ac_parsing.rs`).
- [x] `cargo clippy --manifest-path agents/workflows/feature-task/Cargo.toml --all-targets -- -D warnings` — clean.
- [ ] CI green (requires the PR to be opened so the feature-task workflow gate runs).

## Acceptance Criteria

- [x] AC1: The doing → acceptance gate mechanically verifies cited-file existence, cited-test pass/fail, and evidence-text matching against the PR diff inside the lobster itself, without requiring an agent invocation, and posts the same structured must-fix feedback pattern as existing lobster gates on failure (🧪 testID: agents/workflows/feature-task/src/ac_parsing.rs::tests — 15 new unit tests covering TestId file path in-diff/missing/suffix-match, TestId test name pass/fail, NotTested file path in-diff/missing/free-text, NotCode always-passes, Pr #hash/URL/path, multi-AC isolation, subsection scoping, empty-task-acs short-circuit)
[...]

PR #503, opened 2026-08-21
feat(ash): slim verify.ts to semantic-only; defaultJudgeIntent placeholder (task 5e35dc25 PR #3)

rowanstoffer:
## Summary

Lands the Ash-side cleanup half of task 5e35dc25. PR #502 wired the lobster-side mechanical-evidence gate (cited-file existence, cited-test pass/fail, evidence-text matching against the PR diff) into `verify_delivery` before the `qa_agent` branch. This PR reduces `agents/ash/src/verify.ts` so Ash's qa_agent approval is only requested after the lobster's mechanical gate passes, and is scoped to **semantic judgment of AC intent** — not re-checking facts the lobster already verified.

- Drop the mechanical functions from `verify.ts`: `parseEvidenceTag`, `extractAcEvidence`, `checkTestId`, `checkNotTested`, `checkPrReference`, `checkNotCode`, `checkAcEvidence`, `runAllAcChecks`. These duplicated the lobster's `mechanical_evidence_failures` at `agents/workflows/feature-task/src/ac_parsing.rs:331` and have been the source of two live incidents (missing prodlike credential; `FORGEABLE_APPROVAL_OWNER` bug).
- Add `extractAcLines` — minimal AC walker that strips trailing evidence annotations so the LLM judge prompt sees the bare AC description, not the testID tag.
- Add `defaultJudgeIntent` placeholder. Quinn owns the real implementation (LLM call against `ac.description + pr.patch` with a strict `{ ok: true } | { ok: false; reason: string }` JSON response, deterministically reproducible by `temperature=0` + recorded model+prompt version). The placeholder returns `ok: false` for every AC so tasks stay in `doing` until the real semantic check is wired.
- Rename `verify` → `verifySemantic`. The CLI shape is unchanged (`--task-id`, `--pr-url`, env vars); only the `Deps` type gained `judgeIntent`.
- `defaultFetchPr` now pulls the raw `.patch` text alongside the JSON file list so the judgment function gets the unified diff as prompt context, not just the per-file metadata.
- Update `agents/ash/README.md` to reflect the new scope and `docs/systems/tasks.md` to document the new mechanical-evidence gate ordering.

## Test plan

- [x] `pnpm test` in `agents/ash`: 11 passed, 0 failed (4 semantic-contract pin tests + 4 `extractAcLines` tests + 1 `defaultJudgeIntent` test + 2 pre-conditions tests).
- [x] `npx tsc --noEmit` in `agents/ash`: clean (strict mode, no `any` casts).
- [x] `cargo test --bin feature-task mechanical_evidence` in `agents/workflows/feature-task`: 15 passed, 0 failed (pre-existing tests still pass; this PR does not touch the lobster code).

## Acceptance Criteria

- [x] AC1: The doing → acceptance gate mechanically verifies cited-file existence, cited-test pass/fail, and evidence-text matching against the PR diff inside the lobster itself, without requiring an agent invocation, and posts the same structured must-fix feedback pattern as existing lobster gates on failure (🔗 pr: #502)
- [x] AC2: Ash's qa_agent approval is only requested after the mechanical gate above has passed, and is scoped to semantic judgment of AC intent rather than re-checking facts the lobster already verified (🔗 pr: #502)
- [x] AC3: Ash's verify.ts no longer needs its own service credential/auth principal for the mechanical checks it currently performs — only for granting the final qa_agent approval, if a separate grant step is still required by the design (🧪 testID: agents/ash/test/verify.test.ts)
- [x] AC4: Existing verify.ts test coverage for the mechanical checks is preserved (ported to the lobster's test suite, not dropped) with no regression in what currently gets caught (🔗 pr: #502)

## Follow-up

Quinn implements `defaultJudgeIntent` via the LLM call described in the design's step 4. Until then, the placeholder keeps tasks in `doing` so the gate doesn't transition on a stub. The lobster's mechanical-evidence gate continues to work without Ash running — a task simply never reaches acceptance until Quinn's semantic check exists and is wired.


quinnstoffer (approved):
LGTM. Clean scope split — verify.ts now purely semantic, mechanical checks correctly delegated to the lobster's ac_parsing.rs (already merged in #502). Tests are solid: extractAcLines, defaultJudgeIntent placeholder contract, and the semantic-orchestrator pin tests (all-pass, one-fail, pre-conditions, prompt-shape) all cover real behavior, not just execution. docs/systems/tasks.md update accurately describes the new gate ordering. Noted: the defaultJudgeIntent placeholder returns ok:false for every AC until I implement the real judgeIntent — picking that up as my next action item so this doesn't silently stall the qa_agent gate for every future task.

PR #504, opened 2026-08-21
feat(tasks-api): durable BullMQ+Redis auto-post path in dev and prodlike (task 1945f8a2)

rowanstoffer:
## Summary

The Content Scheduler auto-post worker was running the in-process job adapter in prodlike (no `CONTENT_SCHEDULER_JOB_ADAPTER` / `CONTENT_SCHEDULER_REDIS_URL` set). In-process jobs are held as in-memory `setTimeout`s inside the tasks-api process and are lost on any restart, with no reconciliation safety net because the worker itself never ran. This silently dropped 4 approved / scheduled tweets (2026-08-11 through 2026-08-14) with zero errors — Tom noticed them stuck in the Unscheduled overflow 2026-08-18.

This PR wires the existing BullMQ adapter / worker / reconciliation infrastructure into the dev plane so `make up MODE=prodlike` (and `MODE=dev`) get durable auto-post by default:

- `infra/docker-compose.dev.yml`: add a `redis:7-alpine` service alongside postgres. Shared between modes (project name keeps containers separate). No persistence volume — the reconciliation sweep on every worker boot rebuilds queue state from the database, so a wiped queue is recoverable without manual operator action. The host port publish uses `"${REDIS_PORT}:6379"`, so the actual host port follows the per-mode `REDIS_PORT` (dev=6379, prodlike=6380).
- `scripts/dev/mode-env.sh`: add `REDIS_PORT` to both `dev` (6379) and `prodlike` (6380). The prodlike port is deliberately distinct from dev to match the per-mode port-isolation pattern every other host port follows (POSTGRES_PORT 6432 vs 7432, TASKS_API_PORT 4000 vs 4001, GRAFANA_PORT 3000 vs 3001, etc.) — otherwise running both stacks concurrently on one host would collide on the host port publish even though container names are isolated by compose project.
- `scripts/dev/up.sh`: emit `CONTENT_SCHEDULER_JOB_ADAPTER=bullmq` and `CONTENT_SCHEDULER_REDIS_URL=redis://localhost:${REDIS_PORT}` into the generated tasks-api env file. The BullMQ adapter code already exists in `services/tasks-api/src/routes/contentSchedulerJobs.bullmq.ts`; this PR just turns it on for every mode.
- `scripts/dev/port-cleanup.sh`: also clean `REDIS_PORT` on stale-listener preflight (mirrors the existing Tilt / API / app pattern).
- `infra/tilt/Tiltfile`: add a `redis` `dc_resource` and a `tasks-api-content-scheduler-worker` `local_resource` that runs `npm run content-scheduler:worker` with the same env file as the tasks-api. `resource_deps=['tasks-api', 'redis']` so the worker only starts once both are healthy.

No application code changed — the BullMQ adapter, the worker entrypoint, the reconciliation sweep, and the diagnostics endpoint were all already in place from PR #245 (task `ac74e9bb`). This PR is the missing piece: turning the existing code on in the dev plane.

## Test plan

- [x] `docker compose -f infra/docker-compose.dev.yml config --services` lists `postgres` and `redis`.
- [x] Tiltfile parses cleanly (Python AST).
- [x] `bash -n scripts/dev/up.sh`, `bash -n scripts/dev/port-cleanup.sh`, `bash -n scripts/dev/mode-env.sh` pass.
- [x] `docker compose -f infra/docker-compose.dev.yml --project-name sindustries-verify-1945f8a2 up -d redis` + `redis-cli ping` → `PONG` (verified AC1 infra plumbing).
- [x] MODE=dev: `REDIS_PORT=6379`, `POSTGRES_PORT=6432`, `TASKS_API_PORT=4000`; MODE=prodlike: `REDIS_PORT=6380`, `POSTGRES_PORT=7432`, `TASKS_API_PORT=4001`. `docker compose ... config` confirms host ports: dev publishes redis on 6379, prodlike on 6380 (container port 6379 unchanged).
- [x] Existing test suite `services/tasks-api/test/contentSchedulerAutoPostDurable.test.ts` (BullMQ adapter: 18 tests) + `services/tasks-api/test/contentSchedulerAutoPost.test.ts` (auto-post health endpoint) cover the application code paths this PR turns on. No application code changed in this PR, so no test updates needed.

## Acceptance Criteria

- [x] AC1: Redis is provisioned and reachable from the prodlike environment (local install or hosted instance) (📄 not code: infra/docker-compose.dev.yml adds the redis:7-alpine service; verified with `redis-cli ping` → `PONG` against the freshly-pulled image in this worktree before opening the PR)
- [x] AC2: CONTENT_SCHEDULER_JOB_ADAPTER=bullmq and CONTENT_SCHEDULER_REDIS_URL are set in the tasks-api prodlike env, and the tasks-api process picks them up on restart (📄 not code: scripts/dev/up.sh emits both vars into the generated .env.prodlike and .env.dev; services/tasks-api/src/config/index.ts reads them on next process start)
- [x] AC3: The dedicated auto-post worker (npm run content-scheduler:worker) runs continuously in prodlike under a process manager, not just ad hoc (📄 not code: infra/tilt/Tiltfile adds a tasks-api-content-scheduler-worker local_resource; resource_deps=['tasks-api', 'redis']; Tilt restarts it on file changes and after `make up`)
- [x] AC4: GET /content-scheduler/auto-post/health reports adapter: bullmq, redis reachable, and overdue.count: 0 after the reconciliation sweep runs (⚠️ not tested: the endpoint is unchanged at services/tasks-api/src/routes/contentSchedulerAutoPost.ts and already returns adapter/redis/overdue; this PR turns the env vars on so the next stack restart will populate them. Re-verify on next 'make up MODE=prodlike' with 'curl -sS http://localhost:4001/api/v1/content-scheduler/auto-post/health | jq')
- [x] AC5: A newly approved item scheduled a few minutes in the future auto-publishes without manual intervention, verified end to end in prodlike (⚠️ not tested: wiring complete — worker boots, reconciliation runs, BullMQ scheduler takes new approvals; the actual X publish round-trip requires a real X client plus an approved item scheduled ~2min out. Follow-up manual E2E on the next prodlike stack restart)

## Verification plan for AC4 + AC5 (follow-up commit on next stack restart)

```bash
# AC4: health endpoint reports the right state
curl -sS http://localhost:4001/api/v1/content-scheduler/auto-post/health | jq

# Expect:
# {
#   "data": {
#     "adapter": "bullmq",
#     "redis": { "ok": true, "latencyMs": <small> },
#     "overdue": { "count": 0, "oldestScheduledFor": null, "itemIds": [] },
#     "recommended": null
#   }
# }

# AC5: schedule an item via Mission Control "Unscheduled" overflow
# (or directly via curl) with scheduledFor = now + 2 minutes, approve,
# wait, and verify the tweet is on X with the expected text + that
# /content-scheduler/auto-post/health now shows the published item.
```

## Risk / rollback

- This is infra-only. The BullMQ adapter code path is unchanged. The fallback `in-process` adapter is still selectable by overriding `CONTENT_SCHEDULER_JOB_ADAPTER=in-process` in `.env.prodlike.local` (the local-override grep in `scripts/dev/up.sh` already covers it).
- If Redis crashes in prodlike, the worker exits with a connection error and Tilt restarts it; reconciliation re-enqueues any approved-but-missing jobs on the next boot.
- If a previous in-process job was scheduled before this PR landed and is now lost, the reconciliation sweep on the worker's first boot will re-enqueue it against the BullMQ adapter using the item's `scheduledFor`.
- The prodlike Redis host port is 6380 (not 6379) so dev and prodlike stacks can run concurrently on one host. Operators running only prodlike and not dev should be unaffected — the container-side port is still 6379, only the host publish differs.

## Review feedback addressed

- Quinn's PR #504 deep review (commit a19448d) flagged that the original `mode-env.sh` hardcoded `REDIS_PORT=6379` for both modes, breaking the per-mode port-isolation pattern. Fixed in f3dd798: `REDIS_PORT=6379` for dev and `REDIS_PORT=6380` for prodlike (matching the +1 offset used by TASKS_API_PORT 4000→4001, BUDGET_API_PORT 4002→4003, GRAFANA_PORT 3000→3001).

### DIFF
--- a/infra/docker-compose.dev.yml
+++ b/infra/docker-compose.dev.yml
@@ -17,5 +17,24 @@ services:
       retries: 10
       start_period: 5s
 
+  redis:
+    image: redis:7-alpine
+    restart: unless-stopped
+    ports:
+      - "${REDIS_PORT:-6379}:6379"
+    # Redis is the durable queue backing CONTENT_SCHEDULER_JOB_ADAPTER=bullmq
+    # (see services/tasks-api/src/routes/contentSchedulerJobs.bullmq.ts). The
+    # default config is fine for local dev; persistence is intentionally NOT
+    # enabled here — a wiped queue on `make down` matches dev expectations
+    # (the reconciliation sweep on every worker boot rebuilds queue state
+    # from the database, so a wiped queue is recoverable without manual
+    # operator action).
+    healthcheck:
+      test: ["CMD", "redis-cli", "ping"]
+      interval: 5s
+      timeout: 3s
+      retries: 10
+      start_period: 2s
+
 volumes:
   postgres-data:
--- a/infra/tilt/Tiltfile
+++ b/infra/tilt/Tiltfile
@@ -28,6 +28,12 @@ dc_resource(
   labels=['infra', 'db', mode],
 )
 
+dc_resource(
+  'redis',
+  project_name=compose_project,
+  labels=['infra', 'queue', mode],
+)
+
 local_resource(
   'tasks-api',
   serve_cmd='cd ../../services/tasks-api && DOTENV_CONFIG_PATH=../../{} PORT={} DATABASE_URL="{}" CORS_ALLOWED_ORIGINS="{}" npm run dev'.format(
@@ -69,6 +75,28 @@ local_resource(
   ],
   resource_deps=['postgres'],
   labels=['budget', 'api', mode],
+  allow_parallel=False,
+)
+
+# Content Scheduler auto-post worker (long-lived process; the durable
+# BullMQ adapter side, see services/tasks-api/src/routes/contentSchedulerJobs.bullmq.ts).
+# Runs reconciliation on boot (rebuilds queue state from the database if
+# Redis lost jobs) and then idles until jobs fire. Tilt restart replaces
+# the process; on host restart the worker is brought back by `make up`.
+local_resource(
+  'tasks-api-content-scheduler-worker',
+  serve_cmd='cd ../../services/tasks-api && DOTENV_CONFIG_PATH=../../{} npm run content-scheduler:worker'.format(
+    tasks_api_env_file,
+  ),
+  deps=[
+    '../../services/tasks-api/src',
+    '../../services/tasks-api/prisma',
+    '../../services/tasks-api/package.json',
+    '../../services/tasks-api/package-lock.json',
+    '../../services/tasks-api/.env.example',
+  ],
+  resource_deps=['tasks-api', 'redis'],
+  labels=['tasks', 'worker', mode],
   allow_parallel=False,
 )
 
--- a/scripts/dev/mode-env.sh
+++ b/scripts/dev/mode-env.sh
@@ -14,6 +14,7 @@ case "$MODE" in
     export TASKS_APP_PORT="5173"
     export MISSION_CONTROL_PORT="5176"
     export TILT_PORT="10350"
+    export REDIS_PORT="6379"
     export POSTGRES_DB="sindustries_dev"
     export POSTGRES_CONTAINER_NAME="sindustries-postgres-dev"
     export TASKS_SCHEMA="tasks_api"
@@ -30,9 +31,6 @@ case "$MODE" in
     export TEMPO_PORT="3200"
     export OTLP_GRPC_PORT="4317"
     export OTLP_HTTP_PORT="4318"
-    # Redis (backing CONTENT_SCHEDULER_JOB_ADAPTER=bullmq; see
-    # infra/docker-compose.dev.yml and services/tasks-api/src/routes/contentSchedulerJobs.bullmq.ts).
-    export REDIS_PORT="6379"
     ;;
   prodlike)
     export MODE
@@ -44,6 +42,7 @@ case "$MODE" in
     export TASKS_APP_PORT="5174"
     export MISSION_CONTROL_PORT="5175"
     export TILT_PORT="10351"
+    export REDIS_PORT="6380"
     export POSTGRES_DB="sindustries_prodlike"
     export POSTGRES_CONTAINER_NAME="sindustries-postgres-prodlike"
     export TASKS_SCHEMA="tasks_api"
@@ -60,10 +59,6 @@ case "$MODE" in
     export TEMPO_PORT="3201"
     export OTLP_GRPC_PORT="4327"
     export OTLP_HTTP_PORT="4328"
-    # Redis (backing CONTENT_SCHEDULER_JOB_ADAPTER=bullmq; same port as
-    # dev because docker-compose.dev.yml is shared between modes — only the
-    # project name changes so Redis containers don't actually collide).
-    export REDIS_PORT="6379"
     ;;
   *)
     echo "Unsupported MODE: $MODE (expected: dev | prodlike)" >&2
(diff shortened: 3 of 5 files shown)


quinnstoffer (changes requested):
## Deep Review: durable BullMQ+Redis auto-post path (task 1945f8a2)

Escalated to deep review because this touches infra/tilt/Tiltfile. Verified the "no application code changed" claim — correct, only docker-compose/mode-env/up/port-cleanup/Tiltfile changed.

### Important (should fix before merge)

**`scripts/dev/mode-env.sh:33,63` — `REDIS_PORT` is hardcoded to `6379` for both `dev` and `prodlike`, breaking the per-mode port-isolation pattern every other service follows.**

Every other host port in this file is deliberately differentiated between modes specifically so `dev` and `prodlike` stacks can run concurrently on one host: `POSTGRES_PORT` (6432 vs 7432), `TASKS_API_PORT` (4000 vs 4001), `BUDGET_API_PORT` (4002 vs 4003), `TASKS_APP_PORT`, `MISSION_CONTROL_PORT`, `TILT_PORT`, and all four observability ports (explicitly commented `# Observability ports (per-mode isolation)`). The new `infra/docker-compose.dev.yml` redis service publishes `"${REDIS_PORT:-6379}:6379"` to the host, same as postgres's `"${POSTGRES_PORT:-5432}:5432"` pattern — but unlike postgres, `REDIS_PORT` is the same literal value in both mode branches.

Concretely: if `dev` and `prodlike` are both up (a workflow this whole port scheme exists to support, and one the team actively uses per infra history around the prodlike stack), the second stack's `docker compose up -d redis` will fail to bind host port 6379 (already held by the first stack's redis container) — even though the two stacks are in separate compose projects and don't collide on container names/networks. The PR's own comment in `mode-env.sh` ("Redis containers don't actually collide") is right about containers but doesn't address the host port publish, which does collide.

Second-order effect: `scripts/dev/port-cleanup.sh`'s new `kill_port_listener "${REDIS_PORT:-6379}" "redis"` would then target whichever stack currently holds port 6379 — so `make down MODE=dev` while a `prodlike` redis is live on the same port could kill/interfere with the prodlike stack's redis listener instead of dev's.

**Fix:** give `prodlike` a distinct `REDIS_PORT` (e.g. `6380`), matching the `POSTGRES_PORT` pattern.

### Positive observations

- No application code changed, as claimed — low blast radius otherwise.
- Reconciliation-sweep-on-boot design correctly makes the no-persistence-volume choice safe (wiped queue is recoverable from the DB).
- AC4/AC5 correctly marked as deferred to a real stack restart rather than falsely claimed as verified.
- `resource_deps=['tasks-api', 'redis']` on the new Tilt worker resource is correct ordering.

Everything else checked out (docker-compose config, Tiltfile parse, script syntax, existing BullMQ adapter test coverage). Once `REDIS_PORT` is mode-differentiated, this is good to merge.

quinnstoffer (approved):
Re-approving. The f3dd798 fix correctly differentiates `REDIS_PORT` per mode: 6379 for dev and 6380 for prodlike, matching the +1 offset used by every other per-mode host port (POSTGRES_PORT 6432→7432, TASKS_API_PORT 4000→4001, BUDGET_API_PORT 4002→4003, GRAFANA_PORT 3000→3001). Host-port collision between concurrent dev and prodlike stacks is resolved.

Original concerns addressed:
- mode-env.sh: `REDIS_PORT` is now mode-specific (6379 dev, 6380 prodlike)
- port-cleanup.sh still uses `${REDIS_PORT:-6379}` which resolves per-mode correctly via the env file
- AC4/AC5 still correctly deferred to next stack restart

No further blockers. CI green. Ready to merge.

